PR #16, opened 2025-08-26
add xero integration

lohitya:
Added Xero integration - accounting API to fetch Balance sheet, PL statement, Trial balance, AR and AP statements, find contact id
I have also added unit test, and they are all passing

image


xero/config.json
```diff
@@ -0,0 +1,370 @@
+{
+    "name": "xero",
+    "version": "1.0.0",
```
ProRedCat, line 3: The name of the integration in the config is "xero" but this should be capitalized as "Xero"

xero/xero.py
```diff
@@ -0,0 +1,379 @@
...
+        if not company_name:
+            raise ValueError("company_name is required")
+        if not contact_name:
+            raise ValueError("contact_name is required")
+        
+        try:
+            # Get tenant information by company name
+            tenant_info = await get_tenant_by_company_name(context, company_name)
+            tenant_id = tenant_info["tenant_id"]
+
```
ProRedCat, line 85: We have actions like `get_tenant_by_company_name` but I think this should be changed to not be by company name

I think we should expose a tool which is for getting the available connections (companies) through the https://api.xero.com/connections endpoint
- This way it's possible for the agent to learn of available companies, rather than having to either guess a company name or get told a list of companies
- With this we could then have a required field of Tenant ID for the queries
    - This reduces round trip as we're not constantly doing a request to https://api.xero.com/connections first then doing the query the agent is actually needing
    
    
We also provide the get tenant id by name, so an agent can get the tenant id but they cannot use it

PR #21, opened 2025-09-01
ls/update-xero-integration

lohitya:
- Get Accounts: Retrieve chart of accounts to classify line items (revenue, expenses, assets, etc.) using tenant ID
- Get Payments: Fetch payment records for invoices/bills (customer receipts, supplier payments, refunds) using tenant ID
- Get Bank Transactions: Access bank transactions not tied to invoices (CapEx, financing, operating expenses) using tenant ID

### DIFF
--- a/xero/config.json
+++ b/xero/config.json
@@ -1,5 +1,5 @@
 {
-    "name": "Xero",
+    "name": "xero",
     "version": "1.0.0",
     "description": "Xero accounting integration for aged payables and receivables reporting",
     "entry_point": "xero.py",
--- a/xero/config.json
+++ b/xero/config.json
@@ -1,5 +1,6 @@
 {
     "name": "xero",
+    "name": "Xero",
     "version": "1.0.0",
     "description": "Xero accounting integration for aged payables and receivables reporting",
     "entry_point": "xero.py",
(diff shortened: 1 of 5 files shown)


xero/config.json
```diff
@@ -377,6 +377,106 @@
...
+                        "type": "string",
+                        "description": "Xero tenant ID"
+                    },
+                    "where": {
+                        "type": "string",
+                        "description": "Optional where clause for date filtering"
+                    },
+                    "order": {
+                        "type": "string",
+                        "description": "Optional order parameter for sorting"
```
ProRedCat, line 429: Can we get more detailed descriptions here for the LLM to use. In the README.md file we provide examples on what format is expected

SInce the LLM is effectively writing the filter queries, we should let the LLM know what format Xero expects from this. So hints that this is the Xero where filter (or orderby) and that the format is normally xyz

xero/config.json
```diff
@@ -377,6 +377,106 @@
...
+                    },
+                    "page": {
+                        "type": "integer",
+                        "description": "Optional page number for pagination"
+                    }
```
ProRedCat, line 466: Same comment for these, just giving the LLM more context on how to format the request to Xero

xero/config.json
```diff
@@ -377,6 +377,106 @@
...
+                        "type": "string",
+                        "description": "Xero tenant ID"
+                    },
+                    "where": {
+                        "type": "string",
+                        "description": "Optional where clause for filtering accounts"
+                    },
+                    "order": {
+                        "type": "string",
+                        "description": "Optional order parameter for sorting"
```
ProRedCat, line 397: Last comment here on this one for adding better descriptions for the LLM

ProRedCat (changes requested):
Overall looks good, just added comments in some places for the config which is around better descriptions for the LLM to use the tools

ProRedCat (approved):

PR #59, opened 2025-11-03
fix: update TTS request format to match API specification for speed parameter

lohitya:
This commit implements an update to text to speech request to include speed parameter


TheRealAgentK (changes requested):
The zip file should not be here, should it?

PR #62, opened 2025-11-05
Ls/update front integrattion

lohitya:
This commit fixes the front API channel responses from field 'type' to 'types' (plural) which was causing the Lambda validator to fail when checking the output schema's required properties


TheRealAgentK (approved):
LGTM

PR #63, opened 2025-11-05
Ls/update google docs integration

lohitya:
This commit removed the drive scope from google docs integration. We currently aren't using the Drive scopes within the Google Docs integration


TheRealAgentK (approved):
LGTM

PR #158, opened 2026-01-29
feat(BigQuery): Add Google BigQuery data warehouse integration

Shubhank-Jonnada:
### Description :memo:
- **Purpose**: Add Google BigQuery data warehouse integration to enable serverless data operations, SQL queries, and dataset management
- **Approach**: Implemented comprehensive BigQuery integration with 14 actions using OAuth2 platform authentication and BigQuery REST API v2

**Type of change**
- [ ] Bug fix (non-breaking change which fixes an issue)

**Updates**
👉 Added complete BigQuery integration with 14 actions covering query execution, dataset/table management, data streaming, and job monitoring
👉 Implemented OAuth2 platform authentication with automatic row parsing from BigQuery's nested response format
👉 Added comprehensive documentation, examples, and test suite with support for pagination and all BigQuery data types

### Screenshots :camera:
{Image here of before and after - if applicable}

### Test plan :test_tube:
*Provide guidance for how to QA your proposed changes. This is not only for a test but also useful for a reviewer.*
{A test plan outlining scenarios to test}

- [x] Self-/dev-tested
- [x] Appropriate documentation written (code comments, internal docs)

### DIFF
--- a/bigquery/bigquery.py
+++ b/bigquery/bigquery.py
@@ -9,6 +9,10 @@ bigquery = Integration.load()
 # Base URL for BigQuery API v2
 BIGQUERY_API_BASE = "https://bigquery.googleapis.com/bigquery/v2"
 
+# Maximum number of rows that can be retrieved in a single request
+# This limit protects infrastructure from excessive memory usage
+MAX_RESULTS_LIMIT = 10000
+
 
 def parse_rows(schema: Dict[str, Any], rows: List[Dict[str, Any]]) -> List[Dict[str, Any]]:
     """Convert BigQuery row format to list of dictionaries."""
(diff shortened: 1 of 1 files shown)


CmdrKeen (changes requested):
The code all looks good. 

Codex made the remark that by not specifying the country in the query, the results outside the US data centers would fail. I suggest, if possible, we capture the location information when the integration is connected, and apply it throughout the various queries as they're needed, since it seems like it's more like meta data for the connection.

Since this was initially for an NZ customer, I'm nervous that even if we default to the US, it may appear to be broken for them immediately if they host in Australia, for example. Of course, codex could just be wrong, so I appreciate any feedback too 👍

bigquery/bigquery.py
```diff
@@ -0,0 +1,841 @@
...
+            max_results = inputs.get("max_results", 1000)
+            timeout_ms = inputs.get("timeout_ms", 30000)
+            dry_run = inputs.get("dry_run", False)
+            location = inputs.get("location")
+
+            url = f"{BIGQUERY_API_BASE}/projects/{project_id}/queries"
+
+            payload = {
+                "query": query,
+                "useLegacySql": use_legacy_sql,
```
TheRealAgentK: Is thjs a potential security hole for SQL injection? I'm not sure how BQ deals with this legacy sql internally and that might need to be understood first.

This *might* not be an issue because the user would prob supply the query and the parameters together. @vish4004 - any opinions from an #infosec point of view here?

TheRealAgentK (changes requested):
A few comments to be looked into and fixed, @Sagsgit

bigquery/bigquery.py
TheRealAgentK: I guess it'd be interesting to see how these queries that are typically sent via `query` would be impacted by the `useLegacySql` flag.

TheRealAgentK (approved):

PR #203, opened 2026-04-08
fix(fergus): import HTTPError from module directly

Shubhank-Jonnada:
## Summary

`HTTPError` is not exported from `autohive_integrations_sdk.__init__` on Lambda, causing an `ImportError` at runtime. Importing it directly from `autohive_integrations_sdk.integration` fixes this.

## Changes

- `fergus/fergus.py`: Changed `from autohive_integrations_sdk import HTTPError` to `from autohive_integrations_sdk.integration import HTTPError`

## Testing

Confirmed `from autohive_integrations_sdk.integration import HTTPError` resolves correctly both locally and on Lambda.

### DIFF
--- a/fergus/fergus.py
+++ b/fergus/fergus.py
@@ -5,7 +5,6 @@ from autohive_integrations_sdk import (
     ActionHandler,
     ActionResult,
 )
-from autohive_integrations_sdk.integration import HTTPError
 
 fergus = Integration.load()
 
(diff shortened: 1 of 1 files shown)


TheRealAgentK (approved):
Ok for the time being to unblock you, but we need to have a look why this happens.

TheRealAgentK:
@NinosMan We need to have a look into that, I'm not sure why this is necessary for @Shubhank-Jonnada's integration. 

Edit: To be more precise - it's clear why it's happening - the error *is* in fact not exported in SDK 1.0.2. But is no other integration doing it that way?

Shubhank-Jonnada:
Closing this PR as I’ve reverted the integration to use the inbuilt httperror. With the latest SDK release, we now get detailed error information, which should address the underlying issue behind this PR @TheRealAgentK  @NinosMan

PR #206, opened 2026-04-09
fix(box,circle,clickup): fix ruff lint and bandit issues

Shubhank-Jonnada:
## Summary

Fixes pre-existing lint and security scan failures surfaced by CI across three integrations.

## Changes

| Integration | Issues Fixed |
|-------------|-------------|
| `box` | Added `# nosec B110` to intentional `try/except/pass` in recursive folder listing; added full test suite |
| `circle` | Added `# nosec B105` to placeholder test tokens |
| `clickup` | Fixed 22 `E712` equality comparisons to `True` (e.g. `== True` → truthiness check); added `# nosec B105` to placeholder test tokens |

## Validation

- [x] `ruff check` — clean
- [x] `ruff format` — clean


TheRealAgentK:
@Shubhank-Jonnada `aiohttp` was removed from `box/requirements.txt` in this PR, but `box/box.py` directly imports and uses it:

- **Line 5:** `import aiohttp`
- **Line 202:** `session = aiohttp.ClientSession()`
- **Line 279:** `data = aiohttp.FormData()`
- **Line 290:** `session = aiohttp.ClientSession()`

While `aiohttp` is currently a transitive dependency via the SDK, best practice is to declare direct dependencies explicitly — if the SDK ever changes its dependency tree, box would break silently.

Please add `aiohttp` back to `box/requirements.txt` alongside the SDK pin.

TheRealAgentK:
Also worth considering: why does box use `aiohttp.ClientSession()` and `aiohttp.FormData()` directly instead of using the SDK's built-in HTTP methods? If there's no specific reason, it should be refactored to use the SDK — which would eliminate the direct `aiohttp` dependency entirely. If there is a reason (e.g. multipart form upload not supported by the SDK), please document it.

Shubhank-Jonnada:
Hi @TheRealAgentK, added aiohttp back to box/requirements.txt.

  The direct usage is intentional for two cases that the SDK's context.fetch() doesn't currently support:                                                                                                                                                                                                             
  1. Binary file downloads, context.fetch() calls response.text() internally, which fails for binary content (line 196)
  2. Multipart form uploads, the Box upload API requires multipart form data via aiohttp.FormData() (line 279)

TheRealAgentK (approved):

PR #216, opened 2026-04-09
fix(gitlab,gong,google-ads): fix ruff lint and format issues

Shubhank-Jonnada:
## Summary

- Fixed ruff lint and format issues across `gitlab`, `gong`, and `google-ads` integrations
- Auto-fixed 150 violations using `ruff check --fix --unsafe-fixes`
- Reformatted 11 files using `ruff format`
- Manually wrapped long lines (E501) in `gong/gong.py` and `google-ads/google_ads.py`
- Added `# noqa: E402` to imports that must follow environment variable setup in `google_ads.py`

## Test plan

- [ ] `ruff check` passes with zero errors on all three integrations
- [ ] `ruff format --check` passes on all three integrations


TheRealAgentK:
@Shubhank-Jonnada two things in this one:

**1. gitlab/tests/context.py — broken import**

The `from gitlab import gitlab` line was removed from `context.py`, but the test file does `from context import gitlab` — so that import will now fail. The re-export needs to stay. Please add it back:

```python
from gitlab import gitlab  # noqa: F401
```

**2. google-ads/requirements.txt — please pin the version**

`google-ads==28.4.1` was changed to just `google-ads` with no version constraint. The cleanup of transitive deps (grpcio, protobuf etc.) and adding `proto-plus` is all good, but we should pin google-ads now that we're on v30. Something like `google-ads~=30.0` so we don't accidentally pick up a future major version bump.

Shubhank-Jonnada:
> @Shubhank-Jonnada two things in this one:
> 
> 1. gitlab/tests/context.py — broken import
> 
> The from gitlab import gitlab line was removed from context.py, but the test file does from context import gitlab — so that import will now fail. The re-export needs to stay. Please add it back:
> 
> from gitlab import gitlab  # noqa: F401
> 2. google-ads/requirements.txt — please pin the version
> 
> google-ads==28.4.1 was changed to just google-ads with no version constraint. The cleanup of transitive deps (grpcio, protobuf etc.) and adding proto-plus is all good, but we should pin google-ads now that we're on v30. Something like google-ads~=30.0 so we don't accidentally pick up a future major version bump.
> 

Fixed both of them in https://github.com/Autohive-AI/autohive-integrations/pull/216/commits/7374887a3d6532acbba628ba946a3a5667552a67

TheRealAgentK (approved):

PR #223, opened 2026-04-09
fix(microsoft-word,netlify,notion): fix ruff lint and format issues

Shubhank-Jonnada:
## Summary

- Fixed 190 ruff lint errors across microsoft-word, netlify, and notion integrations
- Applied auto-fixes and unsafe-fixes via `ruff check --fix --unsafe-fixes`
- Applied formatting via `ruff format`
- Manually wrapped remaining E501 long lines in `microsoft-word/microsoft_word.py` using parentheses and variable extraction

## Test plan

- [ ] `ruff check` passes with zero errors on all three integrations
- [ ] `ruff format --check` passes with zero errors on all three integrations


TheRealAgentK (approved):
✅ Approved — lint/format fixes, security annotations (`usedforsecurity=False` on SHA1 in netlify), missing `context.py` added for microsoft-word, icon compressed. Passes `validate_integration` and `check_code`.

PR #228, opened 2026-04-09
fix(whatsapp,X,xero): fix ruff lint and format issues

Shubhank-Jonnada:
## Summary

- Fixed all ruff lint and format issues in the `whatsapp`, `X` (Twitter/X), and `xero` integrations
- Removed duplicate test classes and functions in `xero/tests/test_xero.py` (F811 redefinition errors)
- Wrapped long lines (E501) in `xero/xero.py`, `xero/tests/test_xero.py`, and `X/x.py` to stay within the 120-char limit
- Applied auto-fixes and formatting via `ruff check --fix --unsafe-fixes` and `ruff format`

## Test plan

- [ ] `ruff check` passes with zero errors on `whatsapp/`, `X/`, `xero/`
- [ ] `ruff format --check` passes with zero reformats on all three integrations


TheRealAgentK:
@Shubhank-Jonnada the `X/` → `x/` rename makes sense (lowercase folder names are the convention), but a few things to sort out:

1. **Both `X/` and `x/` exist on the branch** — `X/` still has an `__init__.py` shim that does `from x import x`. This means the tooling will see two integrations. Can we just delete `X/` entirely? Is anything depending on the uppercase path?

2. **Case-insensitive filesystems** — on macOS, `X/` and `x/` are the same directory, which will cause git weirdness for anyone developing on a Mac. Another reason to fully remove `X/` rather than keeping both.

3. This is really more of a `refactor(x): rename X/ to x/` than a lint fix — would've been cleaner as its own commit/PR but not a blocker.

The actual code changes (lint/format, nosec annotations, xero duplicate test removal) all look good. Just need the `X/` directory fully removed.

TheRealAgentK (approved):
Build failure is due to an integration directory being deleted. That is ok in this case.

PR #230, opened 2026-04-10
fix(api-call,app-business-reviews,canva): fix ruff lint and format issues

Shubhank-Jonnada:
## Summary
- ruff format for api-call, app-business-reviews, canva
- nosec B105 on hardcoded token in canva tests


TheRealAgentK (approved):
✅ Approved — lint/format fixes, security annotations, unused variable cleanup, SDK pinning. Passes `validate_integration` and `check_code`.

PR #231, opened 2026-04-10
fix(coda,fergus,jira): fix ruff lint and format issues

Shubhank-Jonnada:
## Summary
- ruff format for coda, fergus, jira
- nosec B105 on hardcoded tokens in coda tests (26 occurrences)
- fergus: added RateLimitError handler in _error() before HTTPError, since RateLimitError is a subclass of HTTPError and must be checked first to be caught correctly (same pattern as typeform)

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Supports basic HTTP authentication and Bearer token authentication via the SDK.
 
 ### Coda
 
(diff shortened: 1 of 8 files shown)


TheRealAgentK:
@Shubhank-Jonnada the fergus change adds a new `RateLimitError` handler in `_error()`, which is a functional addition, not a lint fix. The PR description only mentions "ruff format for coda, fergus, jira" but doesn't call this out.

The change itself looks correct — `RateLimitError` is a subclass of `HTTPError` so it needs to be checked first, and typeform already does the same thing. No objection to the code, just please update the PR description to mention it so the changelog is accurate.

TheRealAgentK (approved):

PR #246, opened 2026-04-15
fix(google-looker): execute_lookml_query API payload

risheetperi:
## Summary

The `execute_lookml_query` action was sending an incorrect payload to the Looker API `POST /queries` endpoint, causing all LookML queries to fail with **HTTP 422: Validation Failed — field `view` is not present**. This blocked querying any dashboard that uses LookML models (including `system__activity` for login security monitoring).

---

## Root Cause

Three field-mapping bugs in the API payload construction:

| Field | Was Sending | Looker API Expects |
|---|---|---|
| explore/view | `"explore": "..."` | `"view": "..."` |
| dimensions + measures | `"dimensions": [...], "measures": [...]` (separate) | `"fields": [...]` (unified array) |
| limit | `100` (integer) | `"100"` (string) |

---

## Changes

- **`google-looker/google_looker.py`** — Fix `ExecuteLookMLQuery` payload: map `explore` input to `view` key, merge `dimensions` + `measures` into single `fields` array, cast `limit` to string
- **`google-looker/tests/context.py`** — Add integration instance re-export so tests can import it
- **`google-looker/tests/test_google_looker_integration.py`** — Add `MockResponse` wrapper matching production fetch shape, add request payload capture, add payload verification assertions (`view` / `fields` / `limit`), add edge-case tests (`dimensions-only`, `measures-only`, `no-fields`), fix result access to use `IntegrationResult` wrapping

> **No changes to `config.json`** — `dimensions` and `measures` remain as user-facing inputs and are merged into `fields` internally.

### DIFF
--- a/google-looker/google_looker.py
+++ b/google-looker/google_looker.py
@@ -167,18 +167,17 @@ class ExecuteLookMLQuery(ActionHandler):
         try:
             helper = build_looker_helper(context)
 
-            query_data = {"model": inputs["model"], "explore": inputs["explore"]}
+            query_data = {"model": inputs["model"], "view": inputs["explore"]}
 
-            if inputs.get("dimensions") is not None:
-                query_data["dimensions"] = inputs.get("dimensions")
-            if inputs.get("measures") is not None:
-                query_data["measures"] = inputs.get("measures")
+            fields = (inputs.get("dimensions") or []) + (inputs.get("measures") or [])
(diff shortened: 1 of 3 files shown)


Shubhank-Jonnada (commented):
lgtm!

Shubhank-Jonnada (approved):

PR #254, opened 2026-04-19
fix(xero): fix attachment 404 errors, add retry logic and input validation

Shubhank-Jonnada:
## Summary

- **Bug fix:** `get_attachments` and `get_attachment_content` now remap `Bills` to `Invoices` before calling the Xero API. Xero has no `/Bills/` endpoint; both sales invoices and purchase bills are accessed via `/Invoices/`. Passing `Bills` previously always returned a 404.
- **Bug fix:** `get_attachment_content` now retries up to 3 times with a 2s delay when Xero returns a 404. This handles Xero's eventual consistency lag where storage is not immediately available right after a file upload.
- **Bug fix:** `attach_file_to_invoice` now validates `name`, `content`, and `contentType` upfront with clear user-facing error messages, and validates that `content` is valid base64 before decoding. Previously these would surface as unhandled exceptions in the error log.
- Removed leftover debug print statements from `attach_file_to_invoice`.
- Config endpoint descriptions updated to document the `Bills` remapping behavior.
- Version bumped to `1.5.0`.

## Test plan

- [ ] Attach a file to a purchase bill, then call `get_attachment_content` with `endpoint: "Bills"` and confirm it returns the file successfully (previously 404)
- [ ] Call `get_attachment_content` immediately after uploading an attachment and confirm it succeeds on retry without manual retrying
- [ ] Call `attach_file_to_invoice` with a missing or empty `contentType` and confirm a clear error message is returned instead of an unhandled exception
- [ ] Call `get_attachments` with `endpoint: "Bills"` and confirm it returns attachment metadata correctly

### DIFF
--- a/xero/xero.py
+++ b/xero/xero.py
@@ -12,6 +12,33 @@ import base64
 
 import aiohttp
 
+
+async def _resolve_file_bytes(file_obj: dict) -> bytes:
+    """Resolve file bytes from a file object that has either 'content' (base64) or 'url' (pre-signed S3)."""
+    content_b64 = file_obj.get("content")
+    if content_b64:
+        try:
+            return base64.b64decode(content_b64)
+        except Exception:
+            raise ValueError("file 'content' is not valid base64-encoded data")
+
+    url = file_obj.get("url")
+    if url:
+        async with aiohttp.ClientSession() as session:
+            async with session.get(url) as resp:
+                if resp.status != 200:
+                    raise ValueError(f"Failed to download file from url: HTTP {resp.status}")
+                return await resp.read()
+
+    file_id = file_obj.get("fileId")
+    if file_id is not None:
+        raise ValueError(
+            f"File content could not be resolved (fileId={file_id}, ownerType={file_obj.get('ownerType')}) — "
+            "the platform failed to inject file content before tool execution"
+        )
+    raise ValueError("file object missing 'content' (base64) or 'url' — ensure a file is attached to the message")
+
+
 # Create the integration using the config.json
 xero = Integration.load()
 
@@ -1333,10 +1360,6 @@ class AttachFileToInvoiceAction(ActionHandler):
                 )
 
             # Extract file data from file object
-            content_b64 = file_obj.get("content")
-            if not content_b64:
-                raise ValueError("file object missing 'content' (expected base64-encoded data)")
-
             file_name = (file_obj.get("name") or "").strip()
             if not file_name:
                 raise ValueError("file object missing 'name'")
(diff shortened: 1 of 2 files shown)


xero/xero.py
```diff
@@ -1341,12 +1335,21 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
...
             content_b64 = file_obj.get("content")
             if not content_b64:
-                raise ValueError("file object missing content")
+                raise ValueError("file object missing 'content' (expected base64-encoded data)")
+
+            file_name = file_obj.get("name", "").strip()
+            if not file_name:
+                raise ValueError("file object missing 'name'")
+
+            content_type = file_obj.get("contentType", "").strip()
```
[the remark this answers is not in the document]
Shubhank-Jonnada: fixed it here: https://github.com/Autohive-AI/autohive-integrations/pull/254/commits/81e5ef64d991133a4303a14a9e15dec6e9c5d548
The fix is (file_obj.get("name") or "").strip() — if the value is None, the or "" kicks in before .strip() is called, so no AttributeError. The validation message then fires correctly as intended.

xero/xero.py
```diff
@@ -1519,8 +1522,11 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
             raise ValueError("guid is required")
 
         try:
+            # Remap Bills -> Invoices (Xero has no /Bills/ endpoint; bills use /Invoices/)
+            api_endpoint = "Invoices" if endpoint == "Bills" else endpoint
```
TheRealAgentK: Why are we doing the remapping in general? Wouldn't it be cleaner to simply not support `Bills` if it doesn't exist?

xero/xero.py
```diff
@@ -1600,23 +1609,45 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
...
-                                "file": {"name": file_name, "content": "", "contentType": ""},
-                                "success": False,
-                                "error": f"Xero API error getting attachment content: {response.status} - {error_text}",
-                            }
-                        )
-
-                    # Read binary content and encode as base64
-                    file_content = await response.read()
-                    content_base64 = base64.b64encode(file_content).decode("utf-8")
+                # Retry with backoff on 404 — Xero storage can lag after upload
```
TheRealAgentK: Is this "lag" documented on Xero's end?

Where do the 3 x 2 sec delay come from? Anecdotal?

Shubhank-Jonnada: It was anecdotal. No Xero docs or community reports back up storage lag as a real cause, 404s on attachments are auth/scope issues. Removed it in the latest commit : https://github.com/Autohive-AI/autohive-integrations/pull/254/commits/7a7cd98d14ae35a466fa9091640f1c1334f812e4

xero/xero.py
Shubhank-Jonnada: removed the remap and dropped Bills from the enum. Description now tells the LLM to use Invoices for both : https://github.com/Autohive-AI/autohive-integrations/pull/254/commits/2ca4be38a06a27382eebe14f989f8804dcde09ea

TheRealAgentK (approved):
looks ok now

PR #255, opened 2026-04-19
fix(google-chat, google-sheets): wrap action returns in ActionResult

Shubhank-Jonnada:
## Summary

- `integration-google-chat` and `integration-google-sheets` were returning plain dicts from all action handler `execute()` methods instead of wrapping them in `ActionResult`.
- Added `ActionResult` to the SDK import in both files.
- Wrapped every `return {...}` in action handlers with `ActionResult(data=..., cost_usd=0.0)`, covering both success and error paths.
- No logic was changed — only the return type is updated to match the pattern used across the rest of the codebase.

### DIFF
--- a/google-sheets/google_sheets.py
+++ b/google-sheets/google_sheets.py
@@ -1,4 +1,4 @@
-from autohive_integrations_sdk import Integration, ExecutionContext, ActionHandler
+from autohive_integrations_sdk import Integration, ExecutionContext, ActionHandler, ActionResult
 from typing import Dict, Any, List
 from googleapiclient.discovery import build
 from googleapiclient.errors import HttpError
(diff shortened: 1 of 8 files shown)


google-sheets/google_sheets.py
```diff
@@ -58,15 +58,14 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
...
-            return response
+            return ActionResult(data=response, cost_usd=0.0)
         except HttpError as e:
-            return {
-                "files": [],
-                "result": False,
-                "error": f"Google Drive API error: {str(e)}",
-            }
+            return ActionResult(
+                data={"files": [], "result": False, "error": f"Google Drive API error: {str(e)}"},
```
NinosMan, line 64: Not blocking but we should refactor the integration so that it returns ActionError's rather than having an error property in the output schema for each action

NinosMan (approved):

google-sheets/google_sheets.py
Shubhank-Jonnada: Thanks for the review @NinosMan ,  Proceeding without ActionError refactor for now, will do a batch change across all integrations after fixing the AWS alerts.

PR #268, opened 2026-04-22
fix(hubspot): use oneOf in get_company_properties output_schema to accept both strings and objects

### LINKED ISSUE #265: fix(hubspot): property actions return inconsistent types vs output schema
## Problem

The three `get_*_properties` actions (`get_company_properties`, `get_deal_properties`, `get_contact_properties`) have an `include_details` flag (default `false`) that changes the return type of the `properties` array:

- `include_details=false` (default): returns `["name1", "name2", ...]` (array of strings)
- `include_details=true`: returns `[{"name": "...", "label": "...", ...}]` (array of objects)

`get_company_properties` has a detailed `output_schema` that declares `items.type: "object"` — the default string path doesn't match this. The other two actions have underspecified schemas (no `items` definition) so they don't trigger validation issues but have the same code pattern.

## Options to evaluate

1. **Fix the code**: Make `include_details=false` return `[{"name": "name1"}, {"name": "name2"}]` instead of bare strings. More consistent, always returns objects. May break existing workflows that expect strings.
2. **Fix the schema**: Remove the `items` constraint from `get_company_properties` to match the other two. Accepts the polymorphic behaviour.
3. **Schema per mode**: Use a looser schema that accepts both strings and objects via `oneOf`.

## Pre-existing

This pattern exists in the original private repo code — it was not introduced by the SDK 2.0 migration.

## Affected actions

- `get_company_properties` (has strict schema — actively mismatched)
- `get_deal_properties` (underspecified schema — no mismatch but same code pattern)
- `get_contact_properties` (underspecified schema — same)

TheRealAgentK:
## Summary

The `get_company_properties` action's `output_schema` declared `properties` array items as strictly `type: "object"`, but when `include_details` is `false` (the default), the Python code returns an array of plain strings (property names). This caused a validation error at runtime.

## Changes

- **`hubspot/config.json`**: Updated the `get_company_properties` output_schema to use `oneOf` for the `properties` array items, accepting both `string` (default) and `object` (when `include_details` is true). Bumped version to `2.0.1`.
- **`hubspot/tests/test_hubspot_companies_unit.py`**: Updated `test_include_details_false` to verify correct data instead of asserting a validation error.

## Validation

- `validate_integration.py` passes (0 errors)
- All 264 unit tests pass

Closes #265

### DIFF
--- a/hubspot/config.json
+++ b/hubspot/config.json
@@ -1,7 +1,7 @@
 {
   "name": "HubSpot",
   "display_name": "HubSpot CRM",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "HubSpot CRM integration for managing contacts, companies, deals, and support tickets with advanced pipeline analytics and UTC date handling",
   "entry_point": "hubspot.py",
   "auth": {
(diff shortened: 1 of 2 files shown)


risheetperi (approved):
LGTM

PR #272, opened 2026-04-27
fix(salesforce): fall back to userinfo endpoint when instance_url missing from credentials

Shubhank-Jonnada:
## Summary

The Salesforce integration was intermittently failing with `instance_url not found in credentials or metadata` across all actions. The platform's OAuth layer does not always forward `instance_url` in the credentials dict — most notably after a token refresh, where only standard OAuth fields are returned. This fix makes the integration self-sufficient: when `instance_url` isn't available through the normal channels, it derives it from Salesforce's `/services/oauth2/userinfo` endpoint using the access token that's always present.

## Root Cause

`_get_token_and_instance` was synchronous and raised immediately if `instance_url` wasn't found in `credentials`, `context.metadata`, or `SALESFORCE_INSTANCE_URL`. After a token refresh the platform reconstructs credentials from the refresh response, which may omit non-standard fields like `instance_url`, causing every action to fail until the user reconnected.

## Changes

- `salesforce/salesforce.py` — made `_get_token_and_instance` async; added `_resolve_instance_url` fallback that calls `https://login.salesforce.com/services/oauth2/userinfo` and parses `urls.rest` to extract the instance hostname; updated all 7 call sites to `await`

## Behaviour

| Scenario | Before | After |
|---|---|---|
| `instance_url` in credentials | Works | Works (unchanged path) |
| `instance_url` in metadata/env | Works | Works (unchanged path) |
| `instance_url` missing (e.g. post-refresh) | Fails with reconnect error | Fetches from userinfo endpoint transparently |

## Notes

- The happy path (instance_url present) has zero extra API calls — the fallback only fires when needed
- Unit tests are unaffected — they set `ctx.metadata["instance_url"]` which still resolves on the second check
- No app-side or platform changes required

### DIFF
--- a/salesforce/config.json
+++ b/salesforce/config.json
@@ -1,7 +1,7 @@
 {
   "name": "Salesforce",
   "display_name": "Salesforce",
-  "version": "1.0.0",
+  "version": "2.0.0",
   "description": "Salesforce CRM integration for searching and updating records, and summarising task and event activity.",
   "entry_point": "salesforce.py",
   "auth": {
--- a/salesforce/salesforce.py
+++ b/salesforce/salesforce.py
@@ -1,4 +1,5 @@
 import re
+from urllib.parse import urlparse
 from autohive_integrations_sdk import (
     Integration,
     ExecutionContext,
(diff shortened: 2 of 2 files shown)


ProRedCat (changes requested):
Instead of fixing this on the integration side, we should instead make it so that the instance url is stored in the metadata field which can then be accessed here

This way we don't have to do a request to fetch what the instance URL is. We have a few integrations that follow this pattern already (see Discord)

Shubhank-Jonnada:
closing this pr:
will address this in the main app

> Instead of fixing this on the integration side, we should instead make it so that the instance url is stored in the metadata field which can then be accessed here
> 
> This way we don't have to do a request to fetch what the instance URL is. We have a few integrations that follow this pattern already (see Discord)

PR #274, opened 2026-04-27
fix(github): return ActionError instead of malformed ActionResult on errors

Shubhank-Jonnada:
## Summary

All GitHub actions use a shared `@handle_github_errors` decorator for error handling. On any exception (including 404s), the decorator was returning `ActionResult(data={"error": ..., "result": False})`. This dict doesn't satisfy the output schema of actions like `list_organization_repositories` (which expects `repositories: array`), causing the SDK to throw a `ValidationError` instead of returning a clean error to the caller.

This has been firing in production since December 2025 (232 instances).

## Root Cause

`ActionError` was not imported and the decorator used `ActionResult` for both success and error paths. The SDK validates `ActionResult.data` against the output schema — an error dict will never match an action's success schema.

## Changes

- `github/github.py` — import `ActionError`; replace both `ActionResult` error returns in the decorator with `ActionError(message=...)` — fixes all actions at once since they all share the decorator

## Before / After

| Scenario | Before | After |
|---|---|---|
| 404 / any exception | `ValidationError` crash | Clean `ActionError` returned to caller |
| No access token | `ValidationError` crash | Clean `ActionError` returned to caller |

### DIFF
--- a/github/config.json
+++ b/github/config.json
@@ -90,7 +90,10 @@
             "type": "string"
           },
           "description": {
-            "type": ["string", "null"]
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "private": {
             "type": "boolean"
@@ -105,7 +108,10 @@
             "type": "string"
           },
           "pushed_at": {
-            "type": ["string", "null"]
+            "type": [
+              "string",
+              "null"
+            ]
           },
           "clone_url": {
             "type": "string"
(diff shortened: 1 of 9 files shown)


ProRedCat (dismissed):

TheRealAgentK:
Can *proper* tests please be added before this gets merged, @Shubhank-Jonnada

TheRealAgentK (approved):
This looks good now.

PR #275, opened 2026-04-28
feat(elevenlabs): upgrade to autohive-integrations-sdk v2 with ActionResult/ActionError

Shubhank-Jonnada:
## Summary

Upgrades the ElevenLabs integration from SDK v1 to v2. All action handlers now return typed `ActionResult` and `ActionError` objects instead of raw dicts with `result: bool` / `error: str` fields, aligning with the v2 SDK contract. Output schemas have been cleaned up to remove the now-redundant `result` and `error` fields.

## Actions

| Action | Description |
|--------|-------------|
| `list_voices` | Returns all available voices with search, filtering, and pagination |
| `get_voice` | Returns metadata about a specific voice |
| `get_voice_settings` | Returns stability, similarity boost, style, and speaker boost settings for a voice |
| `list_history` | Returns a list of all previously generated audio items |
| `download_history_audio` | Downloads a history item's audio file as base64-encoded MP3 |
| `get_user_subscription` | Returns subscription tier, character quota, and current usage |
| `text_to_speech` | Converts text to speech using a selected voice; returns base64-encoded MP3 |

### SDK Upgrade (`requirements.txt`)
- `autohive-integrations-sdk~=1.0.2` → `~=2.0.0`

### Action Handlers (`elevenlabs.py`)
- Imported `ActionResult` and `ActionError` from the v2 SDK
- All success paths return `ActionResult(data={...}, cost_usd=0.0)`
- All error paths return `ActionError(message=str(e))` instead of bare dicts
- Response data accessed via `response.data` (v2 `FetchResponse` object) rather than treating the response as a plain dict

### Config Schema (`config.json`)
- Bumped version `1.0.0` → `2.0.0`
- Removed `result` (boolean) and `error` (string) from every action's `output_schema` — these were SDK v1 envelope fields, not domain data
- Fixed `text_to_speech` output schema `required` to include `file` (was incorrectly `["result"]`)

### Tests
- `test_elevenlabs.py` — updated assertions to use `result.type != ResultType.ACTION_ERROR` and `result.result.data`
- `test_elevenlabs_unit.py` — new comprehensive unit test suite covering all 7 actions with mocked `ctx.fetch` and `aiohttp.ClientSession`; no API key required

## Authentication

- **Type**: Custom (API Key)
- **Fields**: `api_key` (password field) — obtained from ElevenLabs Profile > Developers > API Keys


NinosMan (dismissed):

TheRealAgentK (dismissed):
The unit test setup doesn't match what we've done in other new test integrations and deviates from that skill prescribes. Was the skill used, @Shubhank-Jonnada ?

Also - should the old "manual" tests stay around or can they easily be refactored to a proper integration test?

NinosMan (dismissed):

NinosMan (approved):

PR #280, opened 2026-04-28
feat(supadata): upgrade SDK to 2.0.0, add pytest unit tests

Shubhank-Jonnada:
## Summary

- Upgrades Supadata integration to SDK 2.0.0
- Fixes auth pattern: `context.auth.get("credentials", {}).get("api_key", "")`
- All optional params use `inputs.get()`, required params use `inputs["param"]`
- Uses `response.data` / `response.status` from `FetchResponse`
- Returns `ActionError(message=...)` on errors

## Follow-up commits

- **`20eaea3` refactor(supadata): drop package init to remove test import shim** — replaces the `site.getsitepackages()` / `importlib.spec_from_file_location` shim with a 2-line `sys.path` setup in the local `tests/conftest.py`. Possible because `supadata/__init__.py` is now removed (the validator treats it as optional).
- **`9df0b51` test(supadata): add regression test for unauthorized API key (#316)** — explicit test that mirrors the production 401 shape so future regressions are caught at the unit-test layer.
- **`7ab78b5` test(supadata): tighten validation, missing-key, and timestamp tests** — validation error now checks the message names the offending field; new test pins the empty-string `api_key` fallback (sibling of #316); `_ms_to_timestamp` collapsed into a parametrized table.

### Unit Tests (mocked, CI-safe)
17 passed ✅

### Integration Tests (live Supadata API — verified 2026-05-04)
```
supadata/tests/test_supadata_integration.py::test_get_transcript_live    PASSED
supadata/tests/test_supadata_integration.py::test_get_transcript_mocked  PASSED

2 passed in 4.13s
```

### Coverage
- Combined: **100%** (41/41 lines)
- Unit only: **100%**
- Integration only: 83% (happy path)

### Validation
- validate_integration.py: 0 errors, 1 warning (optional `__init__.py`) ✅
- check_code.py: PASSED ✅

Closes #316


TheRealAgentK (changes requested):
SDK 2.0.0 migration looks clean (correctly fixes the `context.auth` envelope bug too) and the 15 unit tests cover the actions. Local CI verified: validate_integration ✅, check_code ✅, ruff ✅, pytest ✅.

Blocking on one thing: **no `tests/test_supadata_integration.py`**. Supadata uses a custom API-key auth, so adding e2e coverage is low-cost — just one env var (`SUPADATA_API_KEY`) and the `live_context` fixture per the [writing-integration-tests skill](https://github.com/autohive-ai/autohive-integrations/blob/master/.agents/skills/writing-integration-tests/SKILL.md). Please add at minimum a happy-path test against a real transcript before merge.

Also: branch name doesn't follow `<type>/<issue#>/<desc>` per [AGENTS.md](https://github.com/autohive-ai/autohive-integrations/blob/master/AGENTS.md), and no linked issue.

TheRealAgentK:
Hey @Shubhank-Jonnada — quick note on why this PR drops `supadata/__init__.py`, since it makes Supadata the only single-file integration on SDK 2.0 without one. I want to make sure the reasoning is in the PR history rather than just commit messages.

### Why the change

The integration imports the upstream Python SDK with `from supadata import Supadata, SupadataError`. The PyPI package is also called `supadata` — same name as our integration folder. With `supadata/__init__.py` present and the repo root on `sys.path` (which pytest does by default), Python's import machinery finds the empty integration package first and never reaches the real PyPI library.

That's the shadow your original PR worked around with the `site.getsitepackages()` + `importlib.util.spec_from_file_location` block in both test files. Removing the empty `__init__.py` removes the shadow entirely, so we can use plain imports and lean on the repo-wide fixtures from the root [`conftest.py`](https://github.com/Autohive-AI/autohive-integrations/blob/master/conftest.py).

### Why this is OK

The validator's rule is "required for package-style integrations, optional for modular integrations with `actions/`". Supadata is single-file rather than modular, so the warning technically applies — but functionally we're in the same boat as the modular integrations: the integration source isn't loaded via `from <name> import …`, it's loaded by entry-point lookup. There are already 4 modular integrations on master without `__init__.py` (`aws`, `facebook`, `humanitix`, `instagram`) and they're fine.

### Why we didn't pick one of the alternatives

- **Restore `__init__.py` + keep the importlib shim** — gives us back ~50 lines of test boilerplate purely to satisfy a heuristic warning.
- **Re-export `Supadata` from `__init__.py`** — papers over a real package collision and ties our integration's import surface to the upstream SDK's.
- **Rename the integration folder** — breaks the public integration identifier in tooling and deployment.

### Follow-up

I'll file a tooling issue against [autohive-integrations-tooling](https://github.com/autohive-ai/autohive-integrations-tooling) suggesting the validator's "optional `__init__.py`" exception be extended to single-file integrations whose folder name collides with a third-party Python package. That'll silence the warning at the right layer rather than us hacking around it here.

Thanks for the original PR — the SDK 2.0 + `ActionError` work is the substance of the fix; everything I added on top is just polishing the test setup and pinning the regression for #316. 🙏

TheRealAgentK (approved):
LGTM now and we've combined it with fixing #316

PR #334, opened 2026-05-22
fix(github): harden connected-account and list_commits against Unhandled Lambda crash

Tram-Nguyen87:
## Summary

This PR closes known fragility holes that can produce the `Unhandled` Lambda error seen in Raygun (`integration-github:14`). The reviewer noted the original Raygun instance was triggered inside `list_commits`, not the connected-account path — so this PR covers both, plus a defensive widening of the action decorator. **Updated after review feedback from @TheRealAgentK to add deterministic page-limit controls and observability.**

### 1. Connected-account handler (original scope)
`GitHubConnectedAccountHandler.get_account_info` had no error handling. The SDK's `Integration.get_connected_account()` also doesn't catch handler exceptions, so any `HTTPError` (revoked/expired token, 5xx outage) propagated out of the Lambda.

**Fix:** wrap `GitHubAPI.get_user(context)` in `try/except`, log a warning, and return an empty `ConnectedAccountInfo` so the Lambda returns normally.

### 2. `list_commits` / `get_commit` null-author crash
The result-shaping code accessed `commit["commit"]["author"]["name"]` directly. GitHub returns `commit.author = null` for commits whose email isn't tied to a GitHub user (deleted accounts, bot commits), which raised `TypeError` mid-action.

**Fix:** new `_commit_summary` helper uses `.get()` chains and treats null author/committer as `{name: null, email: null, date: null}`. Output schema updated to allow `date = null` to match.

### 3. `handle_github_errors` decorator — defensive `CancelledError` catch
The decorator's `except Exception` does not catch `asyncio.CancelledError` (a `BaseException`). If cooperative cancellation ever reaches the action coroutine it would surface as `Unhandled`.

**Fix:** explicit `except asyncio.CancelledError` returns a clean `ActionError`. **Comment rewritten per @TheRealAgentK's suggestion** — this catch does *not* claim to catch Lambda hard timeouts (those kill the process at the OS level, no Python exception fires). It only handles cooperative cancellation reaching the coroutine. Kept as defensive instrumentation; the new `logger.warning` (section 5) will let us verify in production whether this path ever fires.

### 4. Deterministic page-limit on `paginated_fetch` (added per review)

@TheRealAgentK identified the likely upstream cause: `paginated_fetch` had a `while True` loop that ran until GitHub returned a short page. On a large repo, `list_commits` could keep paging until the Lambda runtime cancelled the task — making the `CancelledError` catch reactive rather than preventative.

**Fix:**
- `paginated_fetch` now takes `max_pages: int = 10` and raises `TimeoutError` with guidance to narrow the request (e.g. `sha`, `path`, `since`, `until`) when exceeded.
- `get_commits` forwards `max_pages` to `paginated_fetch`.
- `ListCommits` action reads `max_pages` from inputs (default 10).
- `config.json` exposes `max_pages` as an optional input so callers can raise the cap when they actually need more than 1000 commits.

### 5. Logging (added per review)

@TheRealAgentK asked for *"more details to see what's really going on."* Added two `logger.warning` calls:
- When `paginated_fetch` hits `max_pages` — logs URL, cap value, and items collected.
- When the `CancelledError` catch fires — logs the action name (so we can verify Kai's hypothesis that this path rarely fires in production).

## Caveat on root cause attribution

The actual Python traceback for the original production `Unhandled` events is **not available** — the .NET wrapper that talks to the Lambda only forwards the AWS `FunctionError: Unhandled` header to Raygun, not the `LogResult` payload that contains the Python traceback. The most likely mechanism (unbounded pagination → Lambda runtime cancellation) is now addressed preventatively by section 4. The other fixes each close a real, independently demonstrable fragility identified during code review. A follow-up to forward `LogResult` to Raygun would let us tie future incidents to a specific cause faster.

**Known scope gap:** `get_pull_requests` has its own pagination loop (not using `paginated_fetch`) and is still unbounded if the caller doesn't pass `limit`. Same bug class as Kai flagged. Deferred to a follow-up — the new logging will help us detect if it fires in production before then.

## Test plan (verified locally)

- [x] **92 unit tests pass** (82 prior + 10 new across all scopes of this PR)
- [x] New tests added in this PR:
  - **Connected-account fix:** `TestConnectedAccount::test_returns_populated_account_info`, `test_http_error_returns_empty_info_instead_of_crashing`, `test_generic_exception_returns_empty_info`
  - **Null author/committer fix:** `TestListCommits::test_null_author_does_not_crash`, `test_null_committer_does_not_crash`, `TestGetCommit::test_null_author_does_not_crash`
  - **`CancelledError` catch:** `TestListCommits::test_cancelled_error_returns_action_error`
  - **`max_pages` cap (post-review):**
    - `TestListCommits::test_max_pages_cap_stops_unbounded_pagination` — confirms `max_pages: 2` input stops after exactly 2 fetches with the cap message
    - `TestListCommits::test_default_max_pages_is_ten` — confirms default cap fires at 10 fetches when no override is passed
    - `TestListCommits::test_completes_under_cap_without_firing` — regression guard: 3 pages of data with the default cap of 10 must complete naturally and return all 230 items (catches off-by-one bugs where the cap fires one iteration too early and silently truncates results)
- [x] `validate_integration.py github` — passed
- [x] `check_code.py github` — passed (lint, format, bandit, pip-audit, config sync, fetch patterns)

## Production verification (for reviewer / QA)

**For the connected-account path:**
1. Connect a GitHub account normally.
2. Revoke Autohive's access in GitHub → Settings → Applications → Authorized OAuth Apps.
3. Trigger the connected-account refresh (reload Integrations page).
4. **Expected:** integration card stays usable; no new `Unhandled` event in Raygun.

**For the `list_commits` null-author path:**
1. Pick an old open-source repo with commits authored by deleted GitHub accounts.
2. Run `list_commits` on it.
3. **Expected:** returns commits with `author: null` fields rather than crashing.

**For the `max_pages` cap:**
1. Run `list_commits` against a large repo (e.g. `torvalds/linux`) with no filters and `max_pages: 2`.
2. **Expected:** returns an `ActionError` after exactly 2 page fetches with message `"GitHub pagination stopped after 2 pages..."`. CloudWatch shows the corresponding `logger.warning`.
3. Re-run with `since="2026-05-01T00:00:00Z"` (recent date — small enough commit window to fit under the default cap) and no `max_pages` override.
4. **Expected:** completes normally, returns commits since that date.

### DIFF
--- a/github/github.py
+++ b/github/github.py
@@ -2688,11 +2688,18 @@ class GitHubConnectedAccountHandler(ConnectedAccountHandler):
 
         Returns:
             ConnectedAccountInfo with user's email, username, name, avatar, etc.
+            Falls back to an empty ConnectedAccountInfo when the GitHub API call
+            fails (e.g., revoked/expired token, 5xx outage). The SDK does not
+            catch exceptions raised from this handler, so letting one propagate
+            crashes the Lambda with "Unhandled". Auth failures surface to the
+            user the next time they run an action.
         """
-        # Fetch authenticated user info
-        user_data = await GitHubAPI.get_user(context)
+        try:
+            user_data = await GitHubAPI.get_user(context)
+        except Exception as e:
+            context.logger.warning(f"Failed to fetch GitHub account info: {e}")
+            return ConnectedAccountInfo()
 
-        # Parse name into first/last
         name = user_data.get("name", "")
         name_parts = name.split(maxsplit=1) if name else []
 
--- a/github/github.py
+++ b/github/github.py
@@ -26,6 +26,9 @@ from urllib.parse import quote
 from functools import wraps
 import asyncio
 import base64
+import logging
+
+logger = logging.getLogger(__name__)
 
 github = Integration.load()
 
@@ -62,9 +65,14 @@ def handle_github_errors(action_name: str):
                 return await func(self, inputs, context)
 
             except asyncio.CancelledError:
-                # Lambda runtime / async machinery cancelled the task (e.g., timeout).
-                # CancelledError inherits from BaseException, so a bare `except Exception`
-                # would not catch it and the Lambda would crash with "Unhandled".
+                # Defensive: convert Python-side task cancellation into an
+                # ActionError if cancellation reaches the action coroutine.
+                # This does not catch Lambda hard timeouts or caller-side
+                # invocation cancellation.
+                logger.warning(
+                    "CancelledError caught in action %s — cooperative cancellation reached the action coroutine",
+                    action_name,
+                )
                 return ActionError(
                     message=(
                         f"Action '{action_name}' was cancelled before completing. "
@@ -106,15 +114,21 @@ class GitHubAPI:
         url: str,
         params: Dict[str, Any] = None,
         data_key: str = None,
+        max_pages: int = 10,
     ) -> List[Dict[str, Any]]:
         """
         Generic paginated fetch that handles GitHub's pagination automatically.
+
+        Bounded by max_pages to prevent unbounded loops on large datasets
+        (e.g. list_commits on a repo with thousands of commits) from running
+        until the Lambda runtime cancels the task.
 
         Args:
             context: ExecutionContext
             url: API endpoint URL
             params: Query parameters
             data_key: Key to extract from response (e.g., 'workflows', 'workflow_runs')
+            max_pages: Hard cap on pages fetched. Raises TimeoutError when exceeded.
         """
         if params is None:
             params = {}
@@ -124,8 +138,23 @@ class GitHubAPI:
 
         all_items = []
         headers = GitHubAPI.get_headers(context)
+        pages_fetched = 0
         while True:
+            if pages_fetched >= max_pages:
+                logger.warning(
+                    "paginated_fetch hit max_pages cap (url=%s, max_pages=%d, items_collected=%d)",
+                    url,
+                    max_pages,
+                    len(all_items),
+                )
+                raise TimeoutError(
+                    f"GitHub pagination stopped after {max_pages} pages "
+                    f"({len(all_items)} items). Narrow the request with filters "
+                    "(e.g. sha, path, since, until) or raise max_pages."
+                )
+
             fetch_result = await context.fetch(url, params=params, headers=headers)
+            pages_fetched += 1
             response = fetch_result.data
 
             # Extract items from response
@@ -272,8 +301,9 @@ class GitHubAPI:
         path: str = None,
         since: str = None,
         until: str = None,
+        max_pages: int = 10,
     ) -> List[Dict[str, Any]]:
-        """Get commits for a repository"""
+        """Get commits for a repository (bounded by max_pages)"""
         url = f"{GitHubAPI.BASE_URL}/repos/{owner}/{repo}/commits"
         params = {}
 
@@ -286,7 +316,7 @@ class GitHubAPI:
         if until:
             params["until"] = until
 
-        return await GitHubAPI.paginated_fetch(context, url, params)
+        return await GitHubAPI.paginated_fetch(context, url, params, max_pages=max_pages)
 
     @staticmethod
     async def get_commit(context: ExecutionContext, owner: str, repo: str, sha: str) -> Dict[str, Any]:
@@ -1197,6 +1227,7 @@ class ListCommits(ActionHandler):
             path=inputs.get("path"),
             since=inputs.get("since"),
             until=inputs.get("until"),
+            max_pages=inputs.get("max_pages", 10),
         )
 
         return ActionResult(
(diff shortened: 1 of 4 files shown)


ProRedCat (dismissed):
The error did not come from the auth connection, though this does fix another potential issue. The instance that you were investigating was caused by `list_commits` action throwing an unhandled exception

Tram-Nguyen87:
Thanks for catching that the Raygun instance was actually triggered in `list_commits`, not the connected-account path.

I've expanded this PR to cover both. Latest commit (`8a27ede`) adds:
1. Defensive `_commit_summary` helper so a null `author`/`committer` in `commit["commit"]` no longer raises `TypeError`.
2. `handle_github_errors` now also catches `asyncio.CancelledError` (a `BaseException` that bypasses `except Exception`) — most plausible vector for a Lambda timeout returning `Unhandled`.

github/github.py
```diff
@@ -60,6 +61,16 @@ async def wrapper(self, inputs: Dict[str, Any], context: ExecutionContext) -> Ac
 
                 return await func(self, inputs, context)
 
+            except asyncio.CancelledError:
```
TheRealAgentK, line 67: Would this really catch a timeout from asyncio? It has its own TimeoutException from looking at the docs and I'm not sure a timeout would actually be exposed there via asyncio.

TheRealAgentK:
This is directionally good and the null-author/committer handling looks like a real fix for a plausible `list_commits` crash. I’d like one clarification/change before approving: the `asyncio.CancelledError` path doesn’t seem like it actually proves or fixes Lambda timeout behavior. 

Lambda hard timeouts won’t normally raise a catchable Python `CancelledError`, and `asyncio.wait_for` timeout would surface as `TimeoutError`, already covered by `except Exception`. 

I’d try to add a more deterministic mitigation for long paginated calls, e.g. page/limit controls or an invocation time budget. The connected-account and commit-shaping fixes look good.

Also, it'd be interesting to log some more details in there to see what's really going on.

github/github.py
TheRealAgentK: As a more specific recommendation, I'd change the comment to this:

```
# Defensive: convert Python-side task cancellation into an
# ActionError if cancellation reaches the action coroutine.
# This does not catch Lambda hard timeouts or caller-side
# invocation cancellation.
```

TheRealAgentK (changes requested):

github/github.py
TheRealAgentK, line 112: One point I'd like to raise:

If we do think this is connected to timeouts, one reason could be that the fetches here are essentially unbound.

Right now GitHub pagination loops until GitHub returns fewer than `per_page` items. For large repos, `list_commits` can keep fetching pages until the Lambda dies. A safer shape would be:

```
    async def paginated_fetch(
        context: ExecutionContext,
        url: str,
        params: Dict[str, Any] = None,
        data_key: str = None,
        max_pages: int = 10,
    ) -> List[Dict[str, Any]]:
        ...
        pages_fetched = 0
        while True:
            if pages_fetched >= max_pages:
                raise TimeoutError(
                    f"GitHub pagination stopped after {max_pages} pages. "
                    "Use filters such as sha, path, since, or until to narrow the request."
                )

            fetch_result = await context.fetch(url, params=params, headers=headers)
            pages_fetched += 1
            ...
```

Then for `list_commits`, either hardcode a conservative max or expose it in `config.json`:

```
commits = await GitHubAPI.get_commits(
    context,
    inputs["owner"],
    inputs["repo"],
    sha=inputs.get("sha"),
    path=inputs.get("path"),
    since=inputs.get("since"),
    until=inputs.get("until"),
    max_pages=inputs.get("max_pages", 10),
)
```

That would actually prevent long unbounded pagination from reaching runtime cancellation.

Tram-Nguyen87:
Thanks Kai — addressed all three in 9b53d99 and f7f58bf.
 
  Rewrote the CancelledError comment to your exact wording. Kept the catch as defensive code since CancelledError is a BaseException and slips past except Exception, but agreed it probably never fires in prod.

  Added max_pages=10 to paginated_fetch — raises a TimeoutError with hints to narrow with sha/path/since/until. Wired through list_commits and exposed in config.json so callers can raise it.

  Logging: added logger.warning at both the cap and the CancelledError catch — should tell us within a few weeks whether that catch ever actually fires.

  92 tests pass, including an off-by-one regression guard for the cap.

Shubhank-Jonnada:
All 48/48 live integration tests passed ✅

Tested with a classic PAT covering: repos, commits, branches, files, diff, issues, pull requests (incl. reviews), webhooks, gists, orgs, workflows, tags, and releases. No failures.

@TheRealAgentK

TheRealAgentK:
Thx @Shubhank-Jonnada and @Tram-Nguyen87 - this looks good now from my point of view.

TheRealAgentK (approved):

TheRealAgentK:
@ProRedCat, Could you please have another look at the PR and/or change the "request changes" so that we can move this along? :)

PR #332, opened 2026-05-18
fix(google-looker): upgrade to SDK 2.0.0 and fix Unhandled Lambda errors

Shubhank-Jonnada:
## Summary

Fixes a production incident affecting **15 users, 52 instances** since April 17 2026 where the `integration-google-looker:4` Lambda was reporting `Unhandled` errors.

## Root Cause

`asyncio.CancelledError` is a `BaseException` (not `Exception`) in Python 3.8+. The Looker integration makes **two** HTTP calls per action (one to authenticate, one for the API), which doubles the exposure to Lambda timeouts and task cancellations. When a cancellation occurred mid-flight, it bypassed all `except Exception` blocks and surfaced as an unhandled crash.

On top of that, the integration was still using the pre-2.0.0 SDK pattern — returning error state inside `ActionResult` rather than using `ActionError`, and including a `result: bool` field in all output schemas that the SDK no longer expects.

### `google_looker.py`
- All 7 actions now return `ActionError(message=...)` on failure instead of `ActionResult(data={"result": False, "error": ...})`
- Removed manual HTTP status checks — SDK 2.0.0 raises `HTTPError` automatically for non-2xx responses
- Let `HTTPError` propagate naturally from `_get_access_token` rather than swallowing it
- Cleaned up redundant `except Exception: raise` in `make_request`

### `config.json`
- Bumped version `1.0.1` → `2.0.0`
- Removed `result` (boolean) and `error` (string) fields from all 7 output schemas — these were SDK 1.x artefacts and no longer apply

### Tests
- Deleted `tests/context.py` (deprecated pattern)
- Added `tests/conftest.py` with standard path setup
- Added `tests/test_google_looker_unit.py` — 36 unit tests covering all 7 actions (`pytest.mark.unit`)
- Rewrote `tests/test_google_looker_integration.py` — 6 live tests against real Looker API (`pytest.mark.integration`)

## Test Plan

- [x] All 36 unit tests pass (`pytest -m unit`)
- [x] All 6 integration tests pass against `raygun.looker.com` (`pytest -m integration`)
  - `list_dashboards` — returns dashboard list, respects `fields` param
  - `get_dashboard` — returns correct dashboard by ID
  - `list_models` — returns model list
  - `get_model` — returns correct model by name
  - `list_connections` — returns connections list
- [x] `validate_integration.py` passes
- [x] `ruff check` + `ruff format` clean

PR #340, opened 2026-05-27
fix(google-calendar, google-looker): address post-merge review comments

Shubhank-Jonnada:
## Summary

Addresses post-merge review comments from #339, plus a missing `.env.example` entry flagged for Google Looker (#332).

### Google Calendar
- **`time_min` semantics**: Corrected description — `timeMin` is an exclusive lower bound on event **end time**, not start time
- **`end_date` convention**: Documented and tested that `end.date` is exclusive for all-day events (single-day event on 2026-06-15 requires `end_date="2026-06-16"`)
- **`.env.example`**: Added missing `GOOGLE_CALENDAR_ACCESS_TOKEN` entry
- **Validator warnings**: Replaced `inputs["key"]` with `inputs.get()` for optional params
- **Version bump**: `1.0.0` → `1.1.0` (minor — new `query` and `timezone` inputs)

### Google Looker
- **`.env.example`**: Added missing `LOOKER_BASE_URL`, `LOOKER_CLIENT_ID`, `LOOKER_CLIENT_SECRET`, `LOOKER_TEST_DASHBOARD_ID`, `LOOKER_TEST_MODEL_NAME`

## Commits

- `841104d` — fix: correct `time_min` description
- `cb9e1a4` — chore: add `GOOGLE_CALENDAR_ACCESS_TOKEN` to `.env.example`
- `1b8c6fb` — chore: bump version `1.0.0` → `1.0.1`
- `3e9bcb8` — fix: document and test `end.date` exclusive convention for all-day events
- `ad520f0` — fix: replace `inputs["key"]` with `inputs.get()` to silence validator warnings
- `3246bec` — chore: bump version `1.0.1` → `1.1.0`
- `ade61b5` — chore: add Google Looker env vars to `.env.example`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -88,6 +88,13 @@
 # OAuth2 access token with scope: https://www.googleapis.com/auth/calendar
 # GOOGLE_CALENDAR_ACCESS_TOKEN=
 
+# -- Google Looker --
+# LOOKER_BASE_URL=
+# LOOKER_CLIENT_ID=
+# LOOKER_CLIENT_SECRET=
+# LOOKER_TEST_DASHBOARD_ID=
+# LOOKER_TEST_MODEL_NAME=
+
 # -- Google Forms --
 # OAuth2 access token with scopes: forms.body, forms.responses.readonly
 # GOOGLE_FORMS_ACCESS_TOKEN=
--- a/.env.example
+++ b/.env.example
@@ -84,6 +84,10 @@
 # GOOGLE_SHEETS_ACCESS_TOKEN=
 # GOOGLE_SHEETS_TEST_SPREADSHEET_ID=
 
+# -- Google Calendar --
+# OAuth2 access token with scope: https://www.googleapis.com/auth/calendar
+# GOOGLE_CALENDAR_ACCESS_TOKEN=
+
 # -- Google Forms --
 # OAuth2 access token with scopes: forms.body, forms.responses.readonly
 # GOOGLE_FORMS_ACCESS_TOKEN=
--- a/google-calendar/README.md
+++ b/google-calendar/README.md
@@ -40,8 +40,8 @@ No additional configuration fields are required as authentication is handled thr
 - **Description:** List events from a specified Google Calendar with optional filtering and pagination
 - **Inputs:**
   - `calendar_id`: Calendar ID to list events from (use 'primary' for user's main calendar)
-  - `time_min`: Filter events after this time (RFC3339 timestamp, optional)
-  - `time_max`: Filter events before this time (RFC3339 timestamp, optional)
+  - `time_min`: Exclusive lower bound on event **end time** (RFC3339, optional). Events whose end time is after this value are included — long events that started before the bound but end after it will still appear.
+  - `time_max`: Exclusive upper bound on event **start time** (RFC3339, optional). Only events that start before this value are returned.
   - `max_results`: Maximum number of events to return (1-2500, optional)
   - `page_token`: Pagination token for retrieving next page (optional)
 - **Outputs:**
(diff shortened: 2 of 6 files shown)


TheRealAgentK:
Thanks for jumping on the post-merge review items here — the `timeMin` docs update and the all-day `end_date` clarification both look like the right fixes, and the version bump makes sense given the new inputs from #339.

One thing I noticed: the validator-warning cleanup around optional inputs is mostly on the right track, but a couple of `None` edge cases slipped through.

Specifically, these timezone defaults still won’t apply if the platform passes an explicit `None`:

```python
timezone = inputs.get("timezone", "UTC")
```

and:

```python
timezone = inputs.get("timezone", event_data.get("start", {}).get("timeZone", "UTC"))
```

`dict.get(key, default)` only uses the default when the key is missing. If `inputs` contains `{"timezone": None}`, this returns `None`, so we can send `timeZone: null` to Google Calendar instead of defaulting to UTC / the existing event timezone.

I think these should be:

```python
timezone = inputs.get("timezone") or "UTC"
```

and:

```python
timezone = inputs.get("timezone") or event_data.get("start", {}).get("timeZone", "UTC")
```

There’s a similar issue in `update_event` attendees handling:

```python
if "attendees" in inputs:
    if inputs.get("attendees"):
        event_data["attendees"] = [{"email": email} for email in inputs["attendees"]]
    else:
        event_data.pop("attendees", None)
```

If the platform passes `{"attendees": None}` for an unmapped optional field, this enters the block and removes existing attendees. Since the action sends a full `PUT`, that could silently clear attendee lists during unrelated event updates.

A safer pattern would be:

```python
if inputs.get("attendees") is not None:
    if inputs["attendees"]:
        event_data["attendees"] = [{"email": email} for email in inputs["attendees"]]
    else:
        event_data.pop("attendees", None)
```

That keeps the intended validator cleanup while preserving the distinction between “not provided” / `None` and an intentional empty list to clear attendees. Otherwise this PR looks like it’s addressing the original docs/config issues well.

TheRealAgentK (commented):
See feedback in stream

Shubhank-Jonnada:
> Thanks for jumping on the post-merge review items here — the `timeMin` docs update and the all-day `end_date` clarification both look like the right fixes, and the version bump makes sense given the new inputs from #339.
> 
> One thing I noticed: the validator-warning cleanup around optional inputs is mostly on the right track, but a couple of `None` edge cases slipped through.
> 
> Specifically, these timezone defaults still won’t apply if the platform passes an explicit `None`:
> 
> ```python
> timezone = inputs.get("timezone", "UTC")
> ```
> 
> and:
> 
> ```python
> timezone = inputs.get("timezone", event_data.get("start", {}).get("timeZone", "UTC"))
> ```
> 
> `dict.get(key, default)` only uses the default when the key is missing. If `inputs` contains `{"timezone": None}`, this returns `None`, so we can send `timeZone: null` to Google Calendar instead of defaulting to UTC / the existing event timezone.
> 
> I think these should be:
> 
> ```python
> timezone = inputs.get("timezone") or "UTC"
> ```
> 
> and:
> 
> ```python
> timezone = inputs.get("timezone") or event_data.get("start", {}).get("timeZone", "UTC")
> ```
> 
> There’s a similar issue in `update_event` attendees handling:
> 
> ```python
> if "attendees" in inputs:
>     if inputs.get("attendees"):
>         event_data["attendees"] = [{"email": email} for email in inputs["attendees"]]
>     else:
>         event_data.pop("attendees", None)
> ```
> 
> If the platform passes `{"attendees": None}` for an unmapped optional field, this enters the block and removes existing attendees. Since the action sends a full `PUT`, that could silently clear attendee lists during unrelated event updates.
> 
> A safer pattern would be:
> 
> ```python
> if inputs.get("attendees") is not None:
>     if inputs["attendees"]:
>         event_data["attendees"] = [{"email": email} for email in inputs["attendees"]]
>     else:
>         event_data.pop("attendees", None)
> ```
> 
> That keeps the intended validator cleanup while preserving the distinction between “not provided” / `None` and an intentional empty list to clear attendees. Otherwise this PR looks like it’s addressing the original docs/config issues well.

Thanks @TheRealAgentK , fixed the above issuse in : https://github.com/Autohive-AI/autohive-integrations/pull/340/commits/40a285286c9694a6b00f11141168b6269821b6f6, please have a look

TheRealAgentK (approved):
LGTM now

PR #341, opened 2026-05-27
fix(trello): use min(ids) for list_cards pagination cursor instead of last element

Shubhank-Jonnada:
## Summary

Addresses the pagination cursor bug flagged by Amp on the merged Trello PR.

`list_cards` was using `cards[-1]["id"]` as the `before` cursor for pagination. Trello's `before`/`since` parameters operate on card creation time encoded in the MongoDB ObjectId — but cards are **not guaranteed to be returned in creation order**. They can come back in board/list position order, so using the last element produces a bad cursor when cards have been manually reordered.

The fix uses `min(ids)` across all card IDs in the page, which gives the lexicographically smallest ObjectId — i.e. the oldest-created card — which is the correct `before` cursor regardless of response order.

## Changes

- **`trello.py`**: replace `cards[-1]["id"]` with `min(ids)` for cursor derivation
- **`config.json`**: remove "newest-first in reverse creation order" assumption from `limit` and `next_before` descriptions
- **`README.md`**: same description fixes
- **`tests/test_trello_unit.py`**: update 3 existing cursor assertions to reflect `min()` behaviour; add `test_cursor_uses_oldest_id_not_last_element` covering the reorder case explicitly
- **`.env.example`**: add missing `TRELLO_API_KEY` and `TRELLO_API_TOKEN`

## Test plan

- [x] 72 unit tests passing
- [x] 19 live integration tests passing


Shubhank-Jonnada:
Tested the new `get_card_attachments` action against the live Trello API, integration test passing ✅

TheRealAgentK:
Thanks for digging into the Trello pagination edge case here, after looking at the Trello paging docs and the historical discussion around `before` / `since`, the `min(ids)` approach makes sense to me for deriving the oldest-created card in a page whose response order may not be chronological.

Two small follow-ups that I think would make this even more robust:

1. **Use realistic Trello/ObjectId-style IDs in the cursor tests**

The new tests are covering the right behavior — cursor should come from the oldest ID, not array position. It would be a bit safer to use realistic fixed-length ObjectId-style IDs in those tests rather than short IDs like `c0` / `c1` / `c2` or shortened pseudo-hex strings.

Since the implementation relies on lexicographic ordering matching ObjectId creation-time ordering, realistic 24-character hex IDs make the test better document the assumption and avoid misleading cases like short string ordering (`c10` vs `c2`).

2. **Attachment output schema may need to allow nullable/variable metadata**

`get_card_attachments` returns Trello’s attachment objects raw, which is good, but the schema currently types `bytes` as only an integer:

```json
"bytes": { "type": "integer" }
```

Trello attachment metadata seems a little loose in practice. Uploaded files usually have a numeric byte size, but URL/link-style or older/external attachments can have missing or `null` file-size metadata, and Atlassian’s own generated docs/examples are not especially precise here either.

If output validation is strict, a successful Trello response containing `"bytes": null` could fail schema validation. It may be safer to make this nullable if supported:

```json
"bytes": {
  "type": ["integer", "null"],
  "description": "File size in bytes when available"
}
```

It might also be worth a quick scan of other attachment fields like `edgeColor`, `mimeType`, and `idMember`, since link-style attachments may not always populate the same metadata as uploaded files.

TheRealAgentK (commented):
Left feedback - of the two items, item 1 is more test hardening than anything else. Item 2 however might cause actual issues and it'd be good to look at.

Shubhank-Jonnada:
> Thanks for digging into the Trello pagination edge case here, after looking at the Trello paging docs and the historical discussion around `before` / `since`, the `min(ids)` approach makes sense to me for deriving the oldest-created card in a page whose response order may not be chronological.
> 
> Two small follow-ups that I think would make this even more robust:
> 
> 1. **Use realistic Trello/ObjectId-style IDs in the cursor tests**
> 
> The new tests are covering the right behavior — cursor should come from the oldest ID, not array position. It would be a bit safer to use realistic fixed-length ObjectId-style IDs in those tests rather than short IDs like `c0` / `c1` / `c2` or shortened pseudo-hex strings.
> 
> Since the implementation relies on lexicographic ordering matching ObjectId creation-time ordering, realistic 24-character hex IDs make the test better document the assumption and avoid misleading cases like short string ordering (`c10` vs `c2`).
> 
> 2. **Attachment output schema may need to allow nullable/variable metadata**
> 
> `get_card_attachments` returns Trello’s attachment objects raw, which is good, but the schema currently types `bytes` as only an integer:
> 
> ```json
> "bytes": { "type": "integer" }
> ```
> 
> Trello attachment metadata seems a little loose in practice. Uploaded files usually have a numeric byte size, but URL/link-style or older/external attachments can have missing or `null` file-size metadata, and Atlassian’s own generated docs/examples are not especially precise here either.
> 
> If output validation is strict, a successful Trello response containing `"bytes": null` could fail schema validation. It may be safer to make this nullable if supported:
> 
> ```json
> "bytes": {
>   "type": ["integer", "null"],
>   "description": "File size in bytes when available"
> }
> ```
> 
> It might also be worth a quick scan of other attachment fields like `edgeColor`, `mimeType`, and `idMember`, since link-style attachments may not always populate the same metadata as uploaded files.

Thanks @TheRealAgentK , fixed the above issuse in : https://github.com/Autohive-AI/autohive-integrations/pull/341/commits/d0d0ac655fb880f4d1f4222c1e846505e634a357, please have a look

TheRealAgentK (approved):
LGTM now, 👍 @Shubhank-Jonnada

PR #343, opened 2026-05-29
fix(github): list_pull_requests uses REST endpoint instead of Search API

Tram-Nguyen87:
## What & why
`list_pull_requests` was failing with a recurring **`HTTP 422 Validation Failed`** on certain repos:

> *"The listed users and repositories cannot be searched either because the resources do not exist or you do not have permission to view them."*

**Root cause:** the action was built on the **Search API** (`GET /search/issues`). The Search API matches `repo:owner/repo` against GitHub's *search index*, which is narrower than live REST access — so it 422s for repos missing from that index, notably **private repos the token can still read via the REST API**. This is a functional/correctness bug, distinct from the recent Unhandled-Lambda crash fixes (different symptom, different root cause), but it lives in the same file and now rides on the same hardening rails.

## Changes
- **Switched to the REST List Pull Requests endpoint** (`GET /repos/{owner}/{repo}/pulls`) — no search-index restriction, so private/unindexed repos work.
- `state` / `sort` / `direction` now sent **natively** (the config `sort` enum values are exactly this endpoint's accepted values → removed the lossy `sort_map` translation).
- `author` / `after` / `before` applied **client-side** (not query qualifiers on this endpoint).
- **Pagination now reuses `GitHubAPI.paginated_fetch`**, so the action inherits the `max_pages` cap (default 10, raises a clean `TimeoutError` + warning log on cap) from the crash-hardening pass — replacing the previous unbounded `while` loop. Added `max_pages` as an action input, mirroring `list_commits`.
- **Early-exit on `limit`:** `paginated_fetch` gained an optional `limit` that stops as soon as enough items are collected (and caps `per_page`), so an unfiltered `limit: N` returns the first N without exhausting pages. (Addresses review feedback.)
- Tests updated for the REST endpoint, the `max_pages` cap, and `limit` early-exit.

## Behavior note for reviewers
`paginated_fetch` now takes an optional `limit` that stops early, so `list_pull_requests` returns the first N without exhausting pages when no client-side filter is set (e.g. `limit: 10` on a large repo no longer raises). With `author`/`after`/`before`, the limit applies to *filtered* results, so we still scan up to `max_pages` then slice. The only case that still surfaces a clean `TimeoutError` is an **unfiltered, no-limit** request (e.g. `state="all"`) on a repo larger than `max_pages × 100` — intended hardening behavior, consistent with `list_commits`.

## Testing
- Local build pipeline green: structure validation (0 errors / 0 warnings), code quality (all checks pass), **97 unit tests pass**.
- Manually verified `list_pull_requests` against a **private** repo (`alyanng/AI-Agent-Benchmarking-Dashboard`): previously would 422 → now returns the PR list (10 PRs, mix of open/closed), no 422.


github/github.py
```diff
@@ -451,56 +451,39 @@ async def get_pull_requests(
...
+        ``after``/``before`` are not query qualifiers on this endpoint, so they
+        are applied client-side after fetching. Pagination is delegated to
+        ``paginated_fetch`` and bounded by ``max_pages`` (raises ``TimeoutError``
+        when exceeded); narrow with ``state``/``author``/``after``/``before`` for
+        full coverage on large repos. ``limit`` caps the returned list.
+        """
+        url = f"{GitHubAPI.BASE_URL}/repos/{owner}/{repo}/pulls"
+        params = {"state": state, "sort": sort, "direction": direction}
+
+        prs = await GitHubAPI.paginated_fetch(context, url, params, max_pages=max_pages)
```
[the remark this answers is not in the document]
Tram-Nguyen87: Good catch — fixed in d19394b.

It is  right: with `limit` set on a repo with ≥ `max_pages * 100` matching PRs, delegating to `paginated_fetch` exhausted all pages and raised the `max_pages` TimeoutError instead of returning the first N — regressing the old early-exit behavior.

Fix: added an optional `limit` to `paginated_fetch` that stops and returns as soon as enough items are collected (and caps `per_page` so we don't over-fetch). `list_pull_requests` passes it through only when there's no client-side filter (`author`/`after`/`before`); with a filter the limit applies to filtered results, so we still scan up to `max_pages` before slicing. Other callers are unaffected (`limit` defaults to `None`).

Added `test_limit_returns_first_n_without_exhausting_pages`: full 100-item pages with `limit: 10, max_pages: 2` → returns 10, no error, fetches once.

github/github.py
```diff
@@ -451,56 +463,46 @@ async def get_pull_requests(
...
+        When no client-side filter is requested the endpoint already returns PRs
+        in the requested order, so ``limit`` is pushed into ``paginated_fetch`` to
+        stop early — a plain ``limit`` (e.g. 10) on a huge repo returns the first
+        N rather than exhausting pages and raising the ``max_pages`` TimeoutError.
+        With a client-side filter we must scan up to ``max_pages`` before slicing.
+        """
+        url = f"{GitHubAPI.BASE_URL}/repos/{owner}/{repo}/pulls"
+        params = {"state": state, "sort": sort, "direction": direction}
+
+        fetch_limit = limit if not (author or after or before) else None
```
TheRealAgentK: There is a problem in this area down to line 505.

The PR adds a good optimization for unfiltered requests:

```
fetch_limit = limit if not (author or after or before) else None
prs = await GitHubAPI.paginated_fetch(..., limit=fetch_limit)
```

So if you call:
```
{
  "limit": 10
}
```

then `fetch_limit` is 10, and `paginated_fetch` stops as soon as it has 10 PRs.

But if you call:

```
{
  "author": "kai",
  "limit": 10
}
```

then `fetch_limit` becomes `None`, because `author` is a client-side filter. That means `paginated_fetch` does not stop early. It fetches pages until either:

- GitHub returns a partial/empty page, or
- `max_pages` is reached

Only after fetching is done does it apply:

```
prs = [pr for pr in prs if pr["user"]["login"].lower() == author_lower]
return prs[:limit]
```

The problem with that is:

Suppose page 1 already contains 10 PRs by kai, and the caller only asked for 10. 

The code still keeps fetching page 2, page 3, ... until the repo ends or `max_pages` fires. On a large repo where every page is full, it can hit the cap and return an error, even though it already had enough matching results to satisfy the request.

github/github.py
```diff
@@ -451,56 +463,46 @@ async def get_pull_requests(
...
+        prs = await GitHubAPI.paginated_fetch(context, url, params, max_pages=max_pages, limit=fetch_limit)
+
         if author:
-            q_parts.append(f"author:{author}")
+            author_lower = author.lower()
+            prs = [pr for pr in prs if pr.get("user", {}).get("login", "").lower() == author_lower]
         if after:
-            q_parts.append(f"created:>={after}")
+            prs = [pr for pr in prs if pr.get("created_at") and pr["created_at"] >= after]
```
TheRealAgentK: I'm not sure if comparing ISO strings directly will work here (and in line 503)

For example, `before="2024-01-01"` excludes PRs created on `2024-01-01T...Z`

The previous Search API qualifier `created:<=2024-01-01` included that date. 

I think you might need to properly parse them as date objects before comparison, or the schema in config should maybe require full UTC timestamps. There are multiple ways to attack this problem, I guess.

TheRealAgentK (changes requested):
See comments in code.

github/github.py
Tram-Nguyen87: Fixed in 7c4543b. You're right — with `author`/`after`/`before` set I was passing `fetch_limit=None`, so a `limit: 10` request would scan to `max_pages` and could raise a spurious `TimeoutError` even when page 1 already had 10 matches.

`paginated_fetch` now takes an optional `filter_fn` predicate that is applied **per page** during pagination, so the `limit` counts *matching* items and stops as soon as enough are collected — even with a client-side filter. `get_pull_requests` builds one combined predicate (author + date bounds) and always passes `limit` through.

One subtlety this surfaced: the last-page check now uses the **raw** page size (before filtering), so a full page that gets entirely filtered out still advances to the next page rather than being mistaken for a short/last page. Added `test_author_filter_with_limit_stops_early` (page 1 of 100 matches, `limit: 10, max_pages: 2` → returns 10, one fetch, no error) and `test_author_filter_thins_pages_then_paginates`.

github/github.py
Tram-Nguyen87: Good catch — fixed in 7c4543b. You're right that raw ISO string comparison broke the inclusive-date semantics: `"2024-01-01T09:00:00Z" <= "2024-01-01"` is `False`, so `before="2024-01-01"` was dropping PRs created during that day, whereas the old Search API `created:<=2024-01-01` included them.

I went with parsing rather than tightening the schema, so existing callers passing bare dates keep working. New `_parse_iso_utc()` helper parses both the bound and each PR's `created_at` into aware UTC datetimes (handles trailing `Z` and naive values). A **bare date** maps to the start of the day for `after` and the **end** of the day (`23:59:59.999999`) for `before`, restoring the inclusive whole-day behavior; a full timestamp is compared as-is. Updated the `after`/`before` config descriptions to document this, and added `test_before_date_is_inclusive_of_whole_day` + `test_after_date_is_inclusive_of_whole_day`.

TheRealAgentK:
Small follow-up added in `4f82958`: invalid `after` / `before` date filters now fail fast with an `ActionError` instead of being silently ignored and returning unfiltered PRs.

Also added regression coverage for both invalid `after` and invalid `before` inputs, including asserting that we don’t call GitHub at all when the filter value is invalid.

Verified locally:
- `github/tests/test_github_prs_unit.py` → 25 passed
- `github/tests` → 103 passed

TheRealAgentK (approved):

TheRealAgentK:
@Tram-Nguyen87 if you're happy with my additions, fine for you to merge and get it deployed for testing.

Tram-Nguyen87:
I am happy with the changes on the GitHub integration, PR's merged. Publishing the new version now. Thanks for
  the work on this.

PR #344, opened 2026-05-29
feat(dropbox): upgrade SDK to 2.0.0, fix upload_file input shape, add pytest unit + integration tests

ProRedCat:
## Summary

Upgrades the Dropbox integration to SDK 2.0.0 and fixes a long-standing bug where `upload_file` could never receive a file from the Autohive platform.

### Bug fix — `upload_file` input shape

The previous schema took a flat `content: string` parameter, but the platform delivers uploaded files as a structured `{ name, content (base64), contentType }` object (matches `box`, `gmail`, `canva`, etc.). As a result, `inputs["content"]` was always empty and uploads failed.

- New schema: `upload_file` takes a `file` object plus an optional folder `path`; the destination filename comes from `file.name`.
- Leading slashes are normalised; trailing slashes are stripped; zero-byte files (`content == ""`) are accepted; invalid base64 returns a clean `ActionError` before any HTTP call.

This is a clean break from the 1.x flat-path shape — no backwards-compat fallback.

### SDK 2.0.0 migration

Followed the `upgrading-sdk-v2` skill end-to-end:

- `requirements.txt` pin: `autohive-integrations-sdk~=1.0.2` → `~=2.0.0`.
- `config.json` version: `1.0.0` → `2.0.0`.
- All `context.fetch()` call sites now access `.data` on the returned `FetchResponse`.
- All error returns converted to `ActionError(message=...)`; `error` and error-only `result` fields removed from every action's output schema.
- `list_folder` now omits `cursor` from the response when the API returns `null` so the output stays valid against the `cursor: string` schema.
- Removed unused `account_info.read` OAuth scope (no action calls account-info endpoints) — flagged by `validate_integration.py`.
- Removed the redundant `list_folder_continue` action — pagination is handled by passing the cursor back into `list_folder`.
- Added `dropbox/__init__.py` for package-style import compatibility.

### Tests

- New `dropbox/tests/conftest.py` overrides `mock_context` with the PlatformOauth2 credentials shape.
- New `dropbox/tests/test_dropbox_unit.py` — 35 tests covering all 7 actions: happy paths, request URL/headers/payload verification, exception → `ActionError`, `upload_file` edge cases (zero-byte content, invalid base64, path normalisation, schema rejection of the flat-content shape).
- New `dropbox/tests/test_dropbox_integration.py` — `live_context` fixture that injects the OAuth `Authorization` header and (per the `writing-integration-tests` skill) raises `HTTPError`/`RateLimitError` on non-2xx so live tests exercise the same `try/except → ActionError` path as production. Read-only `list_folder` coverage plus two `@pytest.mark.destructive` lifecycle tests (upload → metadata → temporary_link → delete; create_folder → upload → copy → move → delete) that scope all writes to a per-run UUID folder under `DROPBOX_TEST_FOLDER` and clean up afterwards.
- Deleted the old `tests/test_dropbox.py` / `tests/context.py` testbed scripts.
- Added `DROPBOX_ACCESS_TOKEN` and `DROPBOX_TEST_FOLDER` to root `.env.example`.

### README

Rewrote `dropbox/README.md` to reflect SDK 2.0 behaviour, the new `upload_file` schema, the removal of `result`/`error` keys, and the new test commands. Added a 2.0.0 version history entry.

## Test plan

- [x] `python -m pytest dropbox/tests/test_dropbox_unit.py -v` — **35 passed**
- [x] `python -m pytest dropbox/tests/test_dropbox_integration.py -m integration -v` — **5 skipped** (no live creds locally)
- [x] `ruff check dropbox` — clean
- [x] `ruff format --config ../autohive-integrations-tooling/ruff.toml dropbox` — formatted
- [x] `python ../autohive-integrations-tooling/scripts/validate_integration.py dropbox` — 0 errors, 0 warnings
- [x] `python ../autohive-integrations-tooling/scripts/check_code.py dropbox` — all sub-checks pass (syntax, imports, JSON, ruff lint+format, bandit, pip-audit, config-code sync, fetch patterns)

## Safe live test command

```bash
pytest dropbox/tests/test_dropbox_integration.py -m "integration and not destructive"
```

Destructive tests mutate real data — only run on a test Dropbox account:

```bash
pytest dropbox/tests/test_dropbox_integration.py -m "integration and destructive"
```


dropbox/config.json
```diff
@@ -195,25 +184,16 @@
                         "default": false
                     }
                 },
-                "required": ["path", "content"]
+                "required": ["file"]
             },
             "output_schema": {
```
TheRealAgentK, line 190: Would be nice to tighten the success output schemas while we're already doing the SDK 2 cleanup. Several actions always return a stable success field (`link`/`metadata`, `file`, `folder`, etc.) but the schema doesn't mark those fields as required. Not necessarily a blocker, but making the schemas precise here would better match the implementation and tests.

dropbox/dropbox.py
```diff
@@ -116,34 +145,39 @@ class UploadFileAction(ActionHandler):
 
     async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
         try:
-            # Build Dropbox-API-Arg header with upload parameters
+            file_obj = inputs.get("file") or {}
```
TheRealAgentK: Since `file` is required by the input schema, can we use `inputs["file"]` here instead of `inputs.get("file")`? The current defensive access leaves a config-code sync warning in the tooling even though this path is being actively changed in this PR.

dropbox/README.md
```diff
@@ -297,19 +250,19 @@ To test the integration:
...
-- **1.0.0** - Initial release with 9 actions
-  - Listing: list_folder, list_folder_continue (2 actions)
-  - Metadata: get_metadata (1 action)
-  - Download: get_temporary_link (1 action)
-  - Write Operations: upload_file, create_folder, delete, move, copy (5 actions)
+- **2.0.0**
+  - Upgraded to `autohive-integrations-sdk~=2.0.0` (uses `FetchResponse` and `ActionError`).
+  - **Breaking**: `upload_file` now takes a structured `file` object (`{ name, content, contentType }`) instead of a flat `content` string, so the platform's file inputs can be wired in directly. The `path` input is now an optional destination folder; the file name comes from `file.name`.
+  - Errors are now returned as `ActionError`. Output schemas no longer carry `result`/`error` keys.
+  - Removed the redundant `list_folder_continue` action — `list_folder` already handles cursor-based pagination.
```
TheRealAgentK, line 263: I'd avoid hard-coding action counts in the README/PR description unless they're central to the change — they tend to drift and become easy review noise.

dropbox/dropbox.py
```diff
@@ -116,34 +145,39 @@ class UploadFileAction(ActionHandler):
...
+                content = base64.b64decode(file_obj["content"], validate=True)
+            except (binascii.Error, ValueError) as e:
+                return ActionError(message=f"File content must be valid base64: {e}")
+
+            dropbox_path = _build_upload_path(inputs.get("path") or "", file_name)
+
             api_arg = {
-                "path": inputs["path"],
+                "path": dropbox_path,
                 "mode": inputs.get("mode", "add"),
```
TheRealAgentK, line 166: `update` looks problematic here. Dropbox's `WriteMode` shorthand only works for the void modes like `add`/`overwrite`; `update` needs a rev-aware object rather than the plain string that this currently forwards. Could we either remove `update` from the enum for now, or add a revision input and serialize the proper Dropbox update mode shape with test coverage?

TheRealAgentK (commented):
Thanks for the cleanup here — the SDK 2 migration and the new test coverage are a good improvement.

One operational thing I do want to see before merge: please run the live Dropbox integration tests with real credentials, including the destructive lifecycle tests on a test Dropbox account / safe `DROPBOX_TEST_FOLDER`. The unit coverage is good, but this PR changes the real upload request shape and path/header/body handling, so I’d like confirmation from the actual Dropbox API path rather than only skipped live tests.

Suggested command:

```bash
pytest dropbox/tests/test_dropbox_integration.py -m "integration and destructive"
```

dropbox/dropbox.py
Shubhank-Jonnada: Fixed in dc39ef6 - changed to `inputs["file"]` directly since it's required by the schema.

dropbox/config.json
Shubhank-Jonnada: Fixed in 7510cb7 - added `required` arrays to output schemas for `get_temporary_link`, `upload_file`, `create_folder`, `delete`, `move`, and `copy`.

dropbox/README.md
Shubhank-Jonnada: Fixed in 6c0fd47 - removed the action count from both the description paragraph and the version history entry.

dropbox/dropbox.py
Shubhank-Jonnada: Fixed in b6d81ec - removed `update` from the mode enum since Dropbox's `update` WriteMode requires a rev-aware object shape, not a plain string; can be added back properly once we wire in a `revision` input.

TheRealAgentK:
The code changes are ok now - I'm still missing the results of the integration test runs, including the destructive actions.

@Shubhank-Jonnada 
@ProRedCat 

I think after we can approve and merge.

Shubhank-Jonnada:
Tested all 8 actions against the live Dropbox API — all 16 tests pass ✓

| Action | Tests | Result |
|---|---|---|
| `list_folder` | root listing, limit param, nonexistent path error | ✅ |
| `get_metadata` | file/folder metadata, nonexistent path error | ✅ |
| `upload_file` | add mode, overwrite mode, full lifecycle | ✅ |
| `get_temporary_link` | returns valid HTTP link for uploaded file | ✅ |
| `create_folder` | creates folder, duplicate returns error | ✅ |
| `delete` | deletes file, nonexistent path error | ✅ |
| `copy` | copies file to new path, verifies metadata | ✅ |
| `move` | moves file to new path, verifies name | ✅ |

TheRealAgentK:
The root README still describes the old Dropbox action surface:

- list_folder_continue
- upload mode update
- “9 actions”

Current PR removes list_folder_continue, removes update, and has 8 actions. The root catalogue should be updated so public docs match the new breaking SDK v2 contract.

I would really stop advertising certain numbers of "something" in README files, you're chasing your tail with those to keep in sync..

TheRealAgentK:
Also - I have concerns about the integration tests and the last commits/reverts. See Slack discussion.

TheRealAgentK (changes requested):

TheRealAgentK (approved):
Thanks — the root README issue is fixed now. I re-checked the updated Dropbox catalogue entry and it no longer advertises `list_folder_continue`, `update` mode, or the old 9-action count.

Approving from my side.

PR #345, opened 2026-06-03
feat(active-campaign): add ActiveCampaign integration with 7 actions

Shubhank-Jonnada:
## Summary

ActiveCampaign is an email marketing and CRM platform used by marketing teams to manage and send EDM campaigns. This integration provides read-only access to email campaign performance data — open rates, click rates, bounce rates, and contact-level engagement — enabling teams to track and improve their marketing performance over time.

## Actions

| Action | Description |
|--------|-------------|
| `list_campaigns` | List all email campaigns with derived open rate, click rate, and bounce rate |
| `get_campaign` | Get full details and performance metrics for a single campaign |
| `get_campaign_links` | Get tracked links in a campaign with per-URL click counts |
| `list_contacts` | List and search contacts with filters by list, status, and creation date |
| `get_contact` | Retrieve a single contact by ID |
| `list_contact_activities` | Get per-contact engagement history (opens, clicks, events) |
| `list_lists` | List all contact lists and audience segments |

## Authentication

- **Type**: Custom (API Key)
- **Fields**: `api_key` (from Settings → Developer), `api_url` (full API URL e.g. `https://mycompany.api-us1.com` — found under Settings → Developer)

## Derived Metrics

Open rate, click rate, and bounce rate are calculated client-side from raw campaign fields (`uniqueopens`, `uniquelinkclicks`, `hardbounces`, `softbounces`, `send_amt`) since the API returns raw counts only.

## Files Added

- `active-campaign/config.json` — Integration schema and action definitions
- `active-campaign/active_campaign.py` — Action handlers
- `active-campaign/requirements.txt` — Dependencies (`autohive-integrations-sdk~=2.0.0`)
- `active-campaign/README.md` — Setup and usage documentation
- `active-campaign/icon.png` — 512×512 integration icon
- `active-campaign/tests/conftest.py` — Test path setup
- `active-campaign/tests/test_active_campaign_unit.py` — 18 unit tests (mocked)
- `active-campaign/tests/test_active_campaign_integration.py` — 11 integration tests (live API)

## Testing

- [x] `validate_integration.py` — passed (0 errors, 0 warnings)
- [x] `check_code.py` — passed (0 errors)
- [x] Linting (`ruff`) — clean
- [x] 18 unit tests — all passed
- [x] 11 live integration tests — all passed against ActiveCampaign sandbox account
- [x] Main README updated with alphabetical entry


active-campaign/active_campaign.py
```diff
@@ -0,0 +1,243 @@
...
+    async def execute(self, inputs: Dict[str, Any], context: ExecutionContext) -> ActionResult:
+        try:
+            params: Dict[str, Any] = {
+                "limit": inputs.get("limit", 20),
+                "offset": inputs.get("offset", 0),
+            }
+            if inputs.get("status") is not None:
+                params["filters[status]"] = inputs["status"]
+            if inputs.get("type"):
+                params["filters[type]"] = inputs["type"]
```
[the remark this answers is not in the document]
Shubhank-Jonnada: The bot was right, filters[status] and filters[type] silently do nothing on the v3 campaigns endpoint.
 Removed them entirely rather than leaving broken filters in place, here : https://github.com/Autohive-AI/autohive-integrations/pull/345/commits/f0c73a991641a6d3e284c577fa38a49cb3c45087

active-campaign/active_campaign.py
```diff
@@ -0,0 +1,239 @@
...
+    async def execute(self, inputs: Dict[str, Any], context: ExecutionContext) -> ActionResult:
+        try:
+            contact_id = inputs["contact_id"]
+            response = await context.fetch(
+                f"{_base_url(context)}/contacts/{contact_id}",
+                method="GET",
+                headers=_headers(context),
+            )
+            _raise_for_status(response)
+            return ActionResult(data={"result": True, "contact": response.data}, cost_usd=0.0)
```
TheRealAgentK: Nice work on the new integration overall — I noticed one small response-shape issue here. ActiveCampaign’s retrieve-contact endpoint returns a wrapper payload like `{ "contact": { ... } }`, so returning `response.data` under our own `contact` key means callers receive `contact.contact.email` rather than `contact.email`.

Could we unwrap this the same way `get_campaign` does?

```python
data = response.data or {}
contact = data.get("contact", data)
return ActionResult(data={"result": True, "contact": contact}, cost_usd=0.0)
```

It would also be worth tightening the unit test to assert a real field like `data["contact"]["email"] == "bob@example.com"`, since the current `is not None` assertion doesn’t catch the nesting.

active-campaign/active_campaign.py
```diff
@@ -0,0 +1,239 @@
...
+
+def _headers(context: ExecutionContext) -> Dict[str, str]:
+    api_key = context.auth.get("api_key", "")
+    return {"Api-Token": api_key, "Accept": "application/json"}
+
+
+def _raise_for_status(response: Any) -> None:
+    if response.status >= 400:
+        data = response.data or {}
+        msg = data.get("message") or data.get("error") or f"HTTP {response.status}"
```
TheRealAgentK: Tiny robustness improvement: this assumes error responses are always JSON objects. The live test fetch helper can return plain text for non-JSON upstream responses, so a proxy/HTML/text error would hit `'str' object has no attribute 'get'` and hide the useful upstream message.

Could we make this tolerate non-dict payloads?

```python
data = response.data or {}
if isinstance(data, dict):
    msg = data.get("message") or data.get("error") or f"HTTP {response.status}"
else:
    msg = str(data) or f"HTTP {response.status}"
raise ValueError(msg)
```

Not a big structural change, just makes failures easier to diagnose.

active-campaign/README.md
```diff
@@ -0,0 +1,247 @@
...
+
+#### `list_campaigns`
+
+List all email campaigns with performance metrics. Returns derived open rate, click rate, and bounce rate calculated from raw campaign data.
+
+**Inputs**
+
+| Parameter | Type | Required | Description |
+|-----------|------|----------|-------------|
+| `status` | integer | No | Filter by status: 0=draft, 1=scheduled, 2=sending, 3=paused, 4=stopped, 5=completed |
```
TheRealAgentK: I think these `status` / `type` campaign filters need a quick cleanup. They’re documented here, and referenced later in the use-case/troubleshooting copy, but they aren’t present in `config.json` or passed by `list_campaigns`.

I also couldn’t confirm these as supported query params in the current ActiveCampaign v3 `GET /campaigns` docs — the documented query params I found were `orders[sdate]`, `orders[ldate]`, and `filters[seriesid]`. So unless you’ve confirmed these undocumented filters against a real account, I’d suggest removing the README references rather than adding unsupported inputs.

active-campaign/README.md
```diff
@@ -0,0 +1,247 @@
...
+- List all contact lists and audience segments
+
+## Setup & Authentication
+
+ActiveCampaign uses API key authentication. Two values are required:
+
+| Field | Description |
+|-------|-------------|
+| `api_key` | Your ActiveCampaign API key |
+| `account_name` | Your account subdomain (e.g. `mycompany` from `mycompany.api-us1.com`) |
```
TheRealAgentK: Small docs/setup mismatch: the integration config and code expect a full `api_url`, but the README asks users for `account_name`. If someone follows this literally and enters only the account name/subdomain, the code will build an invalid base URL.

Could we update this section to match `config.json` and the `.env.example`, i.e. tell users to copy the full **API URL** from **Settings → Developer** alongside the API key? That should make setup much smoother.

TheRealAgentK (commented):
Thanks for putting this together — the structure is in good shape and the local validation/test checks are green. I left a few small comments, mostly around response shape and README/API-contract alignment, that should make the integration easier for users to consume and set up.

active-campaign/active_campaign.py
Shubhank-Jonnada: The fix is in commit https://github.com/Autohive-AI/autohive-integrations/commit/6782a28 - unwrapped get_contact response so callers get contact.email directly instead of contact.contact.email. Also tightened the unit test to assert on actual field values.

active-campaign/active_campaign.py
Shubhank-Jonnada: Done in https://github.com/Autohive-AI/autohive-integrations/commit/f815ee5  - _raise_for_status now checks isinstance(data, dict) before calling .get(), so plain text or HTML error responses are handled gracefully instead of crashing with 'str' object has no attribute 'get'.

active-campaign/README.md
Shubhank-Jonnada: Removed in https://github.com/Autohive-AI/autohive-integrations/commit/106a307 - confirmed these params silently do nothing against a live account

active-campaign/README.md
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/1934af3 - updated the auth table and setup steps to match config.json

Shubhank-Jonnada:
Thanks for the review @TheRealAgentK ! All points addressed, ready for a re-review

active-campaign/README.md
```diff
@@ -0,0 +1,245 @@
...
+| Metric | Formula |
+|--------|---------|
+| `open_rate` | `unique_opens / sends × 100` |
+| `click_rate` | `unique_link_clicks / sends × 100` |
+| `bounce_rate` | `(hard_bounces + soft_bounces) / sends × 100` |
+
+## EDM Performance Use Case
+
+### Track overall EDM performance
+Use `list_campaigns` with `status: 5` (completed) to pull all sent campaigns and compare `open_rate`, `click_rate`, and `bounce_rate` across sends over time.
```
TheRealAgentK: Small README cleanup: this still tells users to call `list_campaigns` with `status: 5`, but `status` is no longer part of the `list_campaigns` input schema after the cleanup. Could we reword this so it does not suggest an unsupported parameter? For example: `Use list_campaigns to pull campaigns and compare open_rate, click_rate, and bounce_rate across sends over time. If you only need completed campaigns, filter the returned campaign list client-side by status.`

Shubhank-Jonnada: done, changed it here: https://github.com/Autohive-AI/autohive-integrations/pull/345/commits/c0c410e3120f863e863f2d9ff3c7236af9a7112e

TheRealAgentK (approved):
All requested changes addressed, LGTM

PR #346, opened 2026-06-04
fix(xero): harden rate limiter, connected account, and downloads against Unhandled Lambda crash

Tram-Nguyen87:
> **Reviewer: @TheRealAgentK** — this is PR 1 of 2 applying the #334 (github) hardening playbook across the integrations named in the recurring Raygun Unhandled group. Since you reviewed #334, requesting you on this series. PR 2 (google-sheets) is #347. Note: google-analytics also appears in the Raygun group (May 29 event) with a confirmed crash mechanism (SDK 1.x error paths returning ActionResult with error/result fields, same as the google-looker #332 fix), but its fix is inseparable from its pending SDK 2.0 migration, which is owned separately - findings are documented for whoever picks that up.

## Context

The Raygun error `Lambda function {FunctionName} returned error: Unhandled` is a template group that collects crashes from **all** integrations. Events examined so far name `integration-github` (fixed in #334), `integration-google-analytics` (May 29), and `integration-google-sheets` (Jun 4).

**Honest scope note:** no examined Raygun event has been confirmed naming xero yet — this PR is *preventive* hardening of the same crash classes #334 fixed, found by auditing `xero.py` with the same lens. Every vector below is a real, test-reproduced crash path; the integration Lambda is killed at 30s.

## Crash vectors fixed

1. **Connected-account handler had no error handling** (`xero.py` `get_account_info`) — identical to the #334 root cause. The SDK does not catch handler exceptions, so a revoked/expired token or 5xx crashed the Lambda. Now falls back to a placeholder `ConnectedAccountInfo` with a warning log.
2. **Rate limiter slept up to 60s per retry** (3×60s worst case) on 429s — guaranteed death mid-sleep in a 30s Lambda. The old `max_wait_time=60` guard only rejected delays *longer* than the Lambda lifetime. Defaults reduced (delay 60s→5s, cap 60s→10s) and the cap now enforces a **cumulative** sleep budget across retries; over-budget delays raise `XeroRateLimitExceededException` immediately → clean `ActionError`.
3. **Raw aiohttp downloads had no timeout** (aiohttp default is 300s) in `get_invoice_pdf`, `get_attachment_content`, and `_resolve_file_bytes`. Added a shared 25s `ClientTimeout`.
4. **`asyncio.CancelledError` slipped through every `except Exception`** (it is a `BaseException`). `make_request` now converts cooperative cancellation to `TimeoutError` (covers all rate-limited actions); the four raw-IO actions catch it directly. Same comment wording and `logger.warning` observability as #334.
5. **Empty-string required inputs validated *before* the `try` block** (`get_invoice_pdf`, `attach_file_to_invoice`, `attach_file_to_bill`, `get_attachment_content`) — addresses review feedback. Required fields are schema-validated, but empty strings pass the schema (no `minLength`), so a `ValueError("... is required")` raised on an empty string escaped the handler (SDK `execute_action()` only converts `ValidationError`) as an Unhandled crash. Validation now lives inside each `try`, returning a clean `ActionError` via the existing `except Exception` path.

Not needed here: xero has no pagination loops (every action is single-request), so no `max_pages` equivalent.

## Tests & validation

- 16 new regression tests in `tests/test_xero_crash_hardening_unit.py`: connected-account fallback (incl. CancelledError), sleep-budget enforcement (fail-fast, cumulative, happy-path retry, safe default), cancellation conversion (unit + end-to-end through an action), timeout cap pin, and empty-string required-input validation across all four raw-I/O actions (ActionError + "is required", no network call attempted).
- 138/138 unit tests pass; `validate_integration` / `check_code` (ruff, bandit, pip-audit, config sync) all pass.
- config.json version bumped 2.0.0 → 2.0.1 for the version-check CI.

### DIFF
--- a/xero/xero.py
+++ b/xero/xero.py
@@ -838,16 +838,16 @@ class GetInvoicePdfAction(ActionHandler):
         - file: Object containing content (base64), contentType, and name
         - success: Boolean indicating if the download was successful
         """
-        # Validate required inputs
-        tenant_id = inputs["tenant_id"]
-        invoice_id = inputs["invoice_id"]
-
-        if not tenant_id:
-            raise ValueError("tenant_id is required")
-        if not invoice_id:
-            raise ValueError("invoice_id is required")
-
         try:
+            # Validate required inputs
+            tenant_id = inputs["tenant_id"]
+            invoice_id = inputs["invoice_id"]
+
+            if not tenant_id:
+                raise ValueError("tenant_id is required")
+            if not invoice_id:
+                raise ValueError("invoice_id is required")
+
             # Build URL for specific invoice
             url = f"https://api.xero.com/api.xro/2.0/Invoices/{invoice_id}"
 
@@ -1354,16 +1354,16 @@ class AttachFileToInvoiceAction(ActionHandler):
         Optional fields:
         - include_online: Whether to include the attachment in online invoice (default: true)
         """
-        # Validate required inputs
-        tenant_id = inputs["tenant_id"]
-        invoice_id = inputs["invoice_id"]
-
-        if not tenant_id:
-            raise ValueError("tenant_id is required")
-        if not invoice_id:
-            raise ValueError("invoice_id is required")
-
         try:
+            # Validate required inputs
+            tenant_id = inputs["tenant_id"]
+            invoice_id = inputs["invoice_id"]
+
+            if not tenant_id:
+                raise ValueError("tenant_id is required")
+            if not invoice_id:
+                raise ValueError("invoice_id is required")
+
             # Debug: Log what inputs we're receiving
             # Get file object from inputs
             file_obj = inputs.get("file")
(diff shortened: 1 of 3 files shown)


Tram-Nguyen87:
Re the two CI warnings (both advisory, all checks passed):

- **Unused scopes (Structure)**: pre-existing on master — this PR does not touch `config.json` scopes. Happy to address scope cleanup in a separate PR if wanted.
- **Version "consider minor bump"**: heuristic triggered by new functions/classes in the diff (test helpers + internal hardening code). Staying with the **patch** bump (2.0.0 → 2.0.1) deliberately: no new actions, no new config inputs, no user-facing capability — pure bug fix. (Contrast #334, which took a minor bump because it added the `max_pages` input.)

xero/xero.py
TheRealAgentK, line 1: ⚠️ Should-fix: several touched raw-I/O actions still validate empty required string inputs before entering the `try` block (`get_invoice_pdf` around 841-849, `attach_file_to_invoice` around 1357-1364, `attach_file_to_bill` around 1455-1462, and `get_attachment_content` around 1610-1623). Since SDK `execute_action()` only converts `ValidationError`, a `ValueError("tenant_id is required")` raised here would still escape as an unhandled handler exception if the model/user passes an empty string. Missing fields are schema-validated, but empty strings are allowed by the schema.

Suggested fix: move these validation checks inside the `try`, or add `minLength: 1` in the relevant schemas. Since this PR is specifically unhandled-hardening, I’d prefer catching them inside the action and returning `ActionError`.

TheRealAgentK:
💡 Suggestion: PR body says “13 new regression tests,” but the added crash-hardening file contains 12 tests, and local full Xero unit count is 134. Update the PR body before merge.

TheRealAgentK (commented):
Posted one should-fix note for the remaining unhandled empty-string validation path.

TheRealAgentK (approved):
Re-reviewed the latest changes — the previous empty-string validation concern is addressed, the added regression coverage is solid, and the validation/test suite is green.

Nice work, Tram, especially on tightening the hardening coverage around cancellation, rate-limit sleep budgets, raw download timeouts, and connected-account fallback behavior.

PR #289, opened 2026-04-28
fix(google-sheets): upgrade to SDK 2.0.0, fix Lambda crash on error paths

Shubhank-Jonnada:
## Summary

Fixes production Lambda crash: \`integration-google-sheets\` returning \"Unhandled\" error.

**Root cause:** Error paths returning \`ActionResult(data={\"error\": ..., \"result\": False})\` fail output schema validation in SDK 2.0.0, causing unhandled exceptions and Lambda crashes.

**Fix:**
- Convert all error paths to \`ActionError(message=...)\`
- Remove \`\"error\"\` and boolean \`\"result\"\` from all output schemas
- Upgrade SDK to \`~=2.0.0\`
- Fix optional param access: \`inputs.get(\"pageSize\")\` and \`inputs.get(\"pageToken\")\` instead of bracket access
- Add 51 pytest unit tests
- Add full integration tests covering all 10 actions

### Unit Tests (mocked, CI-safe)

| Integration | Tests | Coverage | Status |
|---|---|---|---|
| google-sheets | 51/51 | 99% | ✅ PASSED |

### Integration Tests (live Google Sheets/Drive API)

| Test Class | Action | Status |
|---|---|---|
| \`TestListSpreadsheets\` | \`sheets_list_spreadsheets\` | ✅ PASSED |
| \`TestGetSpreadsheet\` | \`sheets_get_spreadsheet\` | ✅ PASSED |
| \`TestListSheets\` | \`sheets_list_sheets\` | ✅ PASSED |
| \`TestReadRange\` | \`sheets_read_range\` | ✅ PASSED |
| \`TestWriteRange\` | \`sheets_write_range\` | ✅ PASSED |
| \`TestAppendRows\` | \`sheets_append_rows\` | ✅ PASSED |
| \`TestFormatRange\` | \`sheets_format_range\` | ✅ PASSED |
| \`TestFreeze\` | \`sheets_freeze\` | ✅ PASSED |
| \`TestBatchUpdate\` | \`sheets_batch_update\` | ✅ PASSED |
| \`TestDuplicateSpreadsheet\` | \`sheets_duplicate_spreadsheet\` | ⏭️ SKIPPED |
| **Total** | **13/14** | ✅ |

> \`duplicate_spreadsheet\` skips — token lacks Drive copy permissions for the test file. The action code is correct, it is a permissions constraint on the test token only.

### Validation

| Check | Result |
|---|---|
| \`validate_integration.py\` | ✅ Passed (1 pre-existing scope warning) |
[...]

PR #347, opened 2026-06-04
fix(google-sheets): add HTTP timeout and cancellation handling against Unhandled Lambda crash

Tram-Nguyen87:
> **Reviewer: @TheRealAgentK** — **PR 2 of 2** in the Unhandled-Lambda hardening series (PR 1: #346 xero). This is the integration named in the **2026-06-04 production crash** (Raygun Unhandled group). Note: google-analytics also appears in the group (May 29 event) with a confirmed crash mechanism, but its fix is inseparable from its pending SDK 2.0 migration, which is owned separately - findings are documented for whoever picks that up.

## Context

The Raygun error `Lambda function {FunctionName} returned error: Unhandled` is a template group collecting crashes from all integrations. The Jun 4 event names `integration-google-sheets` (real customer run). The integration Lambda is killed after 30s; google-sheets had two open vectors that fit the crash signature. (#289 previously fixed a different vector — error-path schema validation.)

**Evidence note:** Raygun only forwards the .NET wrapper stack, not the Python traceback, so the exact cause is inferred — same evidence standard #334 shipped with. Both changes harden real crash paths regardless, and the new warning logs make any recurrence diagnosable from CloudWatch.

## Crash vectors fixed

1. **No timeout on any Google API call.** All actions use the synchronous `googleapiclient`; httplib2 defaults to `timeout=None` (block indefinitely), so a stalled call ran until the 30s Lambda kill. Services are now built with `AuthorizedHttp(creds, http=httplib2.Http(timeout=25))` — stalled calls fail cleanly as `ActionError` before the kill. (`httplib2` added explicitly to requirements since it is now imported directly.)
2. **`asyncio.CancelledError` slipped through every `except Exception`** (it is a `BaseException`). New `handle_cancellation` decorator applied to all 10 actions — mirrors the #334 catch, same comment wording (cooperative cancellation reaching the action coroutine; does not catch Lambda hard timeouts) and `logger.warning` observability.
3. **Deterministic payload guard for the other suspected vector:** `sheets_get_spreadsheet` with `include_grid_data=True` can return an enormous payload on large spreadsheets. The per-socket HTTP timeout caps *stalls*, not total download/parse time, so an unbounded full-grid fetch can still exhaust the 30s budget/memory. It now **requires** bounding `ranges` — an unbounded full-grid request is rejected with an `ActionError` directing the caller to provide A1 `ranges` or use `sheets_read_range`, mirroring GitHub's `max_pages` deterministic-cap pattern. (Addresses @TheRealAgentK's should-fix review note.)

Not needed here: no pagination loops (single-request actions with pageToken passthrough), no connected-account handler exists.

## Tests & validation

- 8 new regression tests in `tests/test_google_sheets_crash_hardening_unit.py`: timeout cap pinned below the Lambda timeout on both sheets/drive services, cancellation → ActionError (two actions), normal-error regression guard, and the `include_grid_data` payload guard (unbounded grid rejected / bounded grid with `ranges` allowed).
- 65/65 unit tests pass; `validate_integration` / `check_code` (ruff, bandit, pip-audit, config sync) all pass.
- config.json version bumped 2.0.0 → 2.0.1 for the version-check CI.

### DIFF
--- a/google-sheets/google_sheets.py
+++ b/google-sheets/google_sheets.py
@@ -6,13 +6,26 @@ from autohive_integrations_sdk import (
     ActionError,
 )
 from typing import Dict, Any, List
+from functools import wraps
+import asyncio
+import logging
+
+import httplib2
+from google_auth_httplib2 import AuthorizedHttp
 from googleapiclient.discovery import build
 from googleapiclient.errors import HttpError
 from google.oauth2.credentials import Credentials
 
+logger = logging.getLogger(__name__)
+
 google_sheets = Integration.load()
 
 SPREADSHEET_MIMETYPE = "application/vnd.google-apps.spreadsheet"
+
+# httplib2's default timeout is None (block indefinitely), but the
+# integration Lambda is killed after 30s — a stalled Google API call must
+# fail cleanly before that, not crash the Lambda with "Unhandled".
+HTTP_TIMEOUT_SECONDS = 25
 
 
 def build_credentials(context: ExecutionContext) -> Credentials:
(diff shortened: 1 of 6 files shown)


Tram-Nguyen87:
Re the two CI warnings (both advisory, all checks passed):

- **Unused scopes (Structure)**: pre-existing on master — this PR does not touch `config.json` scopes.
- **Version "consider major bump — breaking changes detected"**: heuristic false positive. The "removed code" it detected is the internal rewrite of `build_sheets_service`/`build_drive_service` (replacing `credentials=` with a timeout-capped `http=` transport). Nothing public changed: all 10 actions intact with identical input/output schemas, auth and scopes untouched, `config.json` unchanged except the version field. Staying with the **patch** bump (2.0.0 → 2.0.1): pure bug fix, no new or removed capability.

google-sheets/google_sheets.py
```diff
@@ -74,11 +115,19 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
 
 @google_sheets.action("sheets_get_spreadsheet")
 class GetSpreadsheet(ActionHandler):
+    @handle_cancellation("sheets_get_spreadsheet")
     async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
         try:
             service = build_sheets_service(context)
             spreadsheet_id = inputs["spreadsheet_id"]
             include_grid = bool(inputs.get("include_grid_data", False))
+            if include_grid:
```
TheRealAgentK: ⚠️ Should-fix: `include_grid_data=True` is identified in the PR as a suspected oversized-payload vector, but this change only logs a warning. That improves observability, but it does not prevent a huge spreadsheet from still streaming/parsing long enough to hit the Lambda hard timeout or memory pressure despite the httplib2 socket timeout.

Suggested fix: add a deterministic guard in the same spirit as GitHub’s `max_pages`: either reject `include_grid_data=True` with an `ActionError` directing users to `sheets_read_range`, expose/range-limit `ranges`, or otherwise cap the requested grid data. If intentionally deferred, please make that explicit as a known remaining gap.

TheRealAgentK: ⚠️ Should-fix: `include_grid_data=True` is identified in the PR as a suspected oversized-payload vector, but this change only logs a warning. That improves observability, but it does not prevent a huge spreadsheet from still streaming/parsing long enough to hit the Lambda hard timeout or memory pressure despite the httplib2 socket timeout.

Suggested fix: add a deterministic guard in the same spirit as GitHub’s `max_pages`: either reject `include_grid_data=True` with an `ActionError` directing users to `sheets_read_range`, expose/range-limit `ranges`, or otherwise cap the requested grid data. If intentionally deferred, please make that explicit as a known remaining gap.

TheRealAgentK:
💡 Suggestion: PR body says “7 new regression tests,” but the added crash-hardening file contains 8 tests. Update the PR body before merge.

TheRealAgentK (commented):
Posted one should-fix note on the remaining oversized `include_grid_data` path.

TheRealAgentK (approved):
@Tram-Nguyen87 I pushed a fix for a leftover docs inaccuracy in README into your PR and happy to approve now.

PR #348, opened 2026-06-05
fix(lumin-pdf): fix pagination, signer normalization, download encoding, and agreement response — v1.2.0

Shubhank-Jonnada:
## Summary
image

- **Pagination fix** — `list_templates` and `list_workspace_members` now omit `page`/`limit` when neither is provided; send both together when either is given; clamp `limit` to the nearest API-accepted value (10, 25, or 50)
- **Signer normalization** — `send_from_template` normalizes `email` → `email_address` and `role` → `signer_role` as required by the Lumin API; added note that this action requires lumin-type templates with defined signer roles
- **Download encoding** — added `Accept: application/json` to `download_signed_document`, `download_agreement`, and `generate_document_from_template` to prevent binary PDF responses that caused UTF-8 decode errors
- **Agreement response unwrapping** — `create_agreement` now correctly unwraps the `{"agreement": {...}}` envelope from the API response
- **Tests** — 55 unit tests all passing; integration tests fixed for `agreement_id` extraction and `signer_role` discovery via `get_template`; added missing `tests/context.py`
- **README** — full rewrite with key features, setup, per-action docs with inputs/outputs, error handling table, troubleshooting, and version history
- **Config** — bumped version `1.1.0` → `1.2.0`; updated `limit` and `signer_role` descriptions

## Test plan

- [x] `ruff check --fix lumin-pdf` — all checks passed
- [x] `ruff format lumin-pdf` — 7 files unchanged
- [x] `validate_integration.py lumin-pdf` — 0 errors, 0 warnings
- [x] 55 unit tests passing (`pytest lumin-pdf/tests/test_lumin_pdf_unit.py -v`)
- [x] Live API tested during development: list_templates, list_workspace_members, send_from_template, upload_document, create_agreement, download_agreement, generate_document_from_template all confirmed working


lumin-pdf/lumin_pdf.py
```diff
@@ -265,10 +279,18 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext) -> Ac
...
     async def execute(self, inputs: Dict[str, Any], context: ExecutionContext) -> ActionResult:
         try:
+            signers = []
+            for s in inputs["signers"]:
+                signer: Dict[str, Any] = {"email_address": s.get("email_address") or s.get("email", "")}
+                if s.get("name"):
+                    signer["name"] = s["name"]
+                if s.get("signer_role") or s.get("role"):
+                    signer["signer_role"] = s.get("signer_role") or s.get("role")
+                signers.append(signer)
```
TheRealAgentK: This normalization drops supported signer fields such as `verification` because it rebuilds each signer with only `email_address`, `name`, and `signer_role`. Lumin's `SignerWithRole` schema supports per-signer verification, so callers using identity verification would have that silently removed before the API call.

Can we preserve the original signer object while normalizing aliases? For example, start with `signer = dict(s)`, set `email_address` from `email` when needed, set `signer_role` from `role` when needed, and only remove alias keys if the API rejects them. A unit test that preserves a `verification` payload would cover this.

lumin-pdf/config.json
```diff
@@ -217,13 +217,13 @@
...
+      "description": "Send a signature request using a saved lumin-type template (created in Lumin's template builder, not an uploaded PDF). The template must have signer roles defined.",
       "input_schema": {
         "type": "object",
         "properties": {
           "template_id": {"type": "string", "description": "The ID of the template to use"},
           "title": {"type": "string", "description": "Title of the signature request"},
-          "signers": {"type": "array", "description": "List of signers with their roles as defined in the template.", "items": {"type": "object"}},
+          "signers": {"type": "array", "description": "List of signers mapped to template roles. Each signer must have email_address (or email) and signer_role matching a role name from the template (use get_template to discover available roles). All template roles must be covered.", "items": {"type": "object", "properties": {"email_address": {"type": "string"}, "email": {"type": "string"}, "name": {"type": "string"}, "signer_role": {"type": "string", "description": "The role name as defined in the template e.g. 'Employee' or 'Employer'"}}}},
```
TheRealAgentK: The schema/docs still allow signer payloads that the API will reject. Lumin requires signer `email_address` and `name`, but the implementation currently sends `email_address: ""` when neither `email_address` nor `email` is provided, and the unit tests include template signer payloads without an email.

Please validate each signer before `fetch` and return an `ActionError` for missing email/name, then update the signer schema/docs/tests so happy paths use valid signer payloads and there is a negative test for missing email/name.

TheRealAgentK (commented):
Posted two findings from the review: one blocker around preserving supported signer fields during normalization, and one should-fix around signer validation/schema/test coverage.

lumin-pdf/lumin_pdf.py
Shubhank-Jonnada: Preserved full signer objects when normalizing aliases and added validation to reject missing email/name before the  API call - https://github.com/Autohive-AI/autohive-integrations/pull/348/commits/bd23c2fbe05a1442b2244e8576a9bbedb6a7cd65

lumin-pdf/config.json
Shubhank-Jonnada: added signer validation before fetch, returns an error for missing email or name. Updated schema, tests, and
  happy paths to use valid payloads - https://github.com/Autohive-AI/autohive-integrations/pull/348/commits/bd23c2fbe05a1442b2244e8576a9bbedb6a7cd65

TheRealAgentK (approved):
Re-reviewed the latest changes. The prior findings around signer field preservation and signer validation are addressed, and local validation/unit checks passed.

PR #349, opened 2026-06-05
fix(active-campaign): remove required from auth.fields — fixes ValidationError on all 7 actions

Shubhank-Jonnada:
## Summary

All 7 active-campaign actions were failing with:
```
ValidationError: 'api_key' is a required property, 'api_url' is a required property
```

The `required: ["api_key", "api_url"]` array inside `auth.fields` caused the SDK to validate these as required action inputs rather than auth credentials. Removing it matches the pattern used by all other custom-auth integrations (e.g. Heartbeat, Lumin PDF).

**Root cause:** `required` inside `auth.fields` is not the correct place to enforce credential presence — the SDK handles auth separately.

## Test plan

- [x] No `required` array in `auth.fields` — matches pattern of working integrations
- [ ] Re-test all 7 actions after merge: `list_campaigns`, `list_lists`, `list_contacts`, `get_campaign`, `get_campaign_links`, `get_contact`, `list_contact_activities`

### DIFF
--- a/active-campaign/config.json
+++ b/active-campaign/config.json
@@ -22,7 +22,7 @@
                     "label": "API URL",
                     "description": "Your ActiveCampaign API URL (e.g. https://mycompany.api-us1.com) — found under Settings > Developer"
                 }
-            },
+            }
         }
     },
     "actions": {
--- a/active-campaign/config.json
+++ b/active-campaign/config.json
@@ -1,6 +1,6 @@
 {
     "name": "ActiveCampaign",
-    "version": "1.0.0",
+    "version": "1.0.1",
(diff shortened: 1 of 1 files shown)


TheRealAgentK:
So, if this problem slipped through from #345 as the OG PR, it seems to indicate the the tests and/or the tooling have a gap.

While this PR itself is valid as a fix for the issue, can you identify what is lacking or needs to be improved that this slipped through?

TheRealAgentK:
I've investigated why these issues slipped through and there will be some tighenting up of tooling and CICD following. 

The PR itself is fine @Shubhank-Jonnada

TheRealAgentK (approved):

PR #351, opened 2026-06-07
fix(active-campaign): fix auth credential reading and duplicate query params on retry

Shubhank-Jonnada:
## Summary

Two runtime bugs in the ActiveCampaign integration that caused all actions to fail in production despite passing local tests.

## Fixes

**1. Auth credential reading**

The SDK wraps auth fields under a `credentials` key in production — `context.auth` looks like `{"credentials": {"api_key": "...", "api_url": "..."}}`. The integration was reading directly via `context.auth.get("api_url")` which returned `None`, causing `_base_url()` to build URLs like `/api/3/campaigns/1` with no hostname.

Added a `_creds()` helper that unwraps the `credentials` key when present, falling back to the flat shape for local tests — same pattern used in ElevenLabs.

**2. Duplicate query params on retry**

Paginated actions (`list_campaigns`, `list_contacts`, `list_lists`) were encoding params into the URL string via `urlencode`. When the SDK retries a failed request, it re-appends the `params` kwarg to the already-modified URL, resulting in duplicated query strings like `?limit=20&offset=0&limit=20&offset=0&limit=20&offset=0`.

Switched to passing params via the `params=` kwarg and only including `limit`/`offset` when explicitly provided by the caller, so no params means no duplication on retry.

## Validation

- [x] `validate_integration.py` — passed
- [x] `check_code.py` — passed
- [x] Integration tests — 11/11 passed against live API

### DIFF
--- a/active-campaign/active_campaign.py
+++ b/active-campaign/active_campaign.py
@@ -9,13 +9,18 @@ from autohive_integrations_sdk import ActionError, ActionHandler, ActionResult, 
 active_campaign = Integration.load()
 
 
+def _creds(context: ExecutionContext) -> Dict[str, Any]:
+    auth = context.auth
+    return auth.get("credentials", auth)
+
+
 def _base_url(context: ExecutionContext) -> str:
-    api_url = context.auth.get("api_url", "").rstrip("/")
+    api_url = _creds(context).get("api_url", "").rstrip("/")
     return f"{api_url}/api/3"
 
 
 def _headers(context: ExecutionContext) -> Dict[str, str]:
-    api_key = context.auth.get("api_key", "")
+    api_key = _creds(context).get("api_key", "")
     return {"Api-Token": api_key, "Accept": "application/json"}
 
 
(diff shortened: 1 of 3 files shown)


active-campaign/active_campaign.py
```diff
@@ -52,16 +57,17 @@ def _derive_rates(campaign: Dict[str, Any]) -> Dict[str, Any]:
...
+                params["limit"] = inputs["limit"]
+            if inputs.get("offset") is not None:
+                params["offset"] = inputs["offset"]
 
             response = await context.fetch(
                 f"{_base_url(context)}/campaigns",
                 method="GET",
                 headers=_headers(context),
-                params=params,
+                params=params if params else None,
```
TheRealAgentK: This still leaves the duplicate-query retry bug for any call that provides explicit params. The SDK's `ExecutionContext.fetch()` appends `params` into the URL before making the request, then on retry recursively calls `fetch(url, ..., params, ...)` with the already-mutated URL and the same params. So `list_campaigns({"limit": 2})`, filtered `list_contacts`, and `list_lists({"offset": ...})` can still retry as `...?limit=2&limit=2...`. `list_contact_activities` also still passes non-empty params.

Please either fix the SDK retry behavior centrally, or avoid `params=` here by building the query string once in the integration and passing `params=None` on the fetch call. Add a regression test for an explicit-params retry case so this doesn't pass while only the no-input/default path is fixed.

TheRealAgentK (dismissed):
Thanks for fixing the production auth shape. I’m requesting changes because the query-param retry fix only avoids duplication when no params are passed; explicit pagination/filter params can still duplicate on retry via the SDK recursive retry path. See inline comment.

active-campaign/active_campaign.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/pull/351/commits/f80d2e7a754f91b68ca079036f10b1772a92d7a0,  moved all params out of params= kwarg and into the URL via _url() + urlencode.

Shubhank-Jonnada:
@codex review

sumitramanga (approved):

sumitramanga:
## Tests :

## Local Tests with live API :
image

## On the app:
image (1)

PR #352, opened 2026-06-08
fix(lumin-pdf): always default page=1 and limit=10 for list actions

Shubhank-Jonnada:
## Summary

`list_templates` and `list_workspace_members` were failing with a 400 error when called without pagination params because the Lumin API requires `page` as a mandatory query parameter. The v1.2.0 refactor introduced a regression where both `page` and `limit` were only sent when the user explicitly provided them — previously `page` defaulted to `1` unconditionally.

Both actions now always send `page=1` and `limit=10` as defaults, with user-supplied values overriding them.

## Root Cause

Introduced in PR #348 (`sj/lumin-fixes`). The original code used `{"page": int(inputs.get("page", 1))}` (always defaulted). The refactor changed this to only include params when the user provided them, breaking no-arg calls.

## Changes

- `lumin_pdf.py` — `list_workspace_members` and `list_templates` now always include `page` (default `1`) and `limit` (default `10`) in the request params
- `tests/test_lumin_pdf_unit.py` — updated two tests that asserted the old broken behaviour (`page` omitted when not provided) to assert the correct defaults instead

### DIFF
--- a/lumin-pdf/README.md
+++ b/lumin-pdf/README.md
@@ -321,6 +321,7 @@ pytest lumin-pdf/tests/test_lumin_pdf_integration.py -m "integration and destruc
 
 ## Version History
 
+- **v1.2.1** — Fixed `list_templates` and `list_workspace_members` always sending `page=1` and `limit=10` defaults, preventing 400 errors when called without pagination params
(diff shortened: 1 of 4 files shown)


Shubhank-Jonnada:
@codex review

sumitramanga (approved):

PR #353, opened 2026-06-08
test(lumin-pdf): add no-arg live tests for list_templates and list_workspace_members

Shubhank-Jonnada:
## Summary

Follow-up to PR #352. Adds live integration tests that call `list_workspace_members` and `list_templates` with no params (`{}`), guarding against the real API 400 regression that PR #352 fixed.

The existing tests only called these actions with `{"limit": 10}`, which happened to work even before the fix (since limit being present triggered page defaulting). A no-arg call is the actual failure scenario that surfaced in production.

## Changes

- `tests/test_lumin_pdf_integration.py` — added `test_returns_members_no_params` and `test_returns_templates_no_params`

### DIFF
--- a/lumin-pdf/README.md
+++ b/lumin-pdf/README.md
@@ -321,6 +321,7 @@ pytest lumin-pdf/tests/test_lumin_pdf_integration.py -m "integration and destruc
 
 ## Version History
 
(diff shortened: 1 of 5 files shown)


TheRealAgentK (approved):
LGTM, can be merged after @Shubhank-Jonnada confirmed the new integration tests (or the whole set) ran and passed.

Shubhank-Jonnada:
> LGTM, can be merged after @Shubhank-Jonnada confirmed the new integration tests (or the whole set) ran and passed.

All green

PR #354, opened 2026-06-09
feat(x): upgrade X integration to SDK 2.0.0 with unit + integration tests

Tram-Nguyen87:
## What & why

Upgrades the **X (formerly Twitter)** integration from SDK `1.0.2` to `2.0.0`, following the `upgrading-sdk-v2` skill and using the Instagram integration as the reference. Adds full unit + integration test suites (the integration previously had only a placeholder script).

### Source — `x.py`
- **FetchResponse:** access the body via `response.data` at every `context.fetch()` call site — the connected-account handler, the `_upload_media` chunked-upload helper, and all 15 actions.
- **ActionError:** convert all error returns (`ActionResult(data={"result": False, "error": ...})` and `except` blocks) to `ActionError(message=...)`; added `ActionError` to the SDK import.

### Config & deps
- `config.json`: removed the now-unused `error`/`result` output-schema properties and dropped `result` from each `required` list — failure is signalled by `ActionError`. Version bumped.
- `requirements.txt`: pinned `autohive-integrations-sdk~=2.0.0`.

### Tests (replace the old placeholder `test_x.py` + `context.py`)
- **`tests/test_x_unit.py`** — 68 mocked tests (`FetchResponse` + `mock_context` pattern). Covers the connected-account handler, `_upload_media` (image/video-poll/error paths), and **all 15 actions** across happy-path, request-verification, `ActionError`, and edge cases.
- **`tests/test_x_integration.py`** — `live_context` fixture (platform-OAuth header injection, Variant 3) with `real_fetch` returning `FetchResponse`. Read-only tests for the handler + read actions, plus `@pytest.mark.destructive` create/bookmark/repost/follow **lifecycle** tests (each cleans up after itself).
- **`tests/conftest.py`** — `sys.path` setup.
- **`.env.example`** — documented `X_ACCESS_TOKEN` and `X_TEST_TARGET_USER_ID`.

> Note: the `x/` folder ships an `__init__.py` that turns `x` into a package exposing only the integration object, so a plain `import x` can't reach the handler/helper. Both test files use the skill's documented **importlib file-path loader** to load `x.py` directly.

## How to run the integration tests

Set `X_ACCESS_TOKEN` (and optionally `X_TEST_TARGET_USER_ID`) in your local `.env`, then:

**Read-only (safe — use this by default):**
```
pytest x/tests/test_x_integration.py -m "integration and not destructive"
```

**Destructive (writes real posts/bookmarks/reposts to the connected account — run deliberately):**
```
pytest x/tests/test_x_integration.py -m "integration and destructive"
```

## Validation (local)
- `ruff check` + `ruff format` (tooling config) ✅
- 68 unit tests passing ✅
- 18 integration tests collect & skip cleanly without a token ✅
- `validate_integration.py` → **0 errors** ✅
- `check_code.py` → **PASSED** (lint, format, bandit, pip-audit, config-code sync, fetch patterns) ✅


Tram-Nguyen87:
Thanks — good catch, and it pointed at a real gap. Resolved in 749a9c0.

**On the production code path:** the SDK's `context.fetch` raises on non-2xx — `RateLimitError` on 429 and `HTTPError` on any other non-ok status (`integration.py` lines ~498–522) — and only returns a `FetchResponse` for 2xx. So in production a 401/403/429 on search already propagates to the action's `except Exception` block and is returned as an `ActionError`, not a successful `posts: []`.

**The real defect was in the test fixture:** the integration test's `real_fetch` returned a `FetchResponse` for *all* statuses, so it did **not** mirror the SDK's raise-on-non-2xx behaviour. That meant a failed request (token scopes / API tier / rate limit) would have surfaced as an empty successful result in tests, and `_skip_if_action_error` would never trigger — exactly the masquerade you describe, but confined to the test harness.

Fix: `real_fetch` now raises `RateLimitError`/`HTTPError` on non-2xx just like the SDK, so the error paths are exercised faithfully. No change needed in the action code, since the SDK already guarantees non-2xx raises.

x/x.py
```diff
@@ -300,16 +297,15 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
 
             return ActionResult(
                 data={
-                    "posts": response.get("data", []),
-                    "includes": response.get("includes", {}),
-                    "meta": response.get("meta", {}),
-                    "result": True,
+                    "posts": response.data.get("data", []),
```
TheRealAgentK: This still treats an X response with an `errors` body as a successful empty result. That can hide scope, tier, or partial API failures as `posts: []` instead of returning an `ActionError`. Please check for `errors` before building the success result, same as the bookmark and create paths do.

.env.example
```diff
@@ -168,3 +168,10 @@
 # YOUTUBE_TEST_THUMBNAIL_PATH=
 # Video ID owned by the authenticated user (required by moderate_comment, which only works on your own channel)
 # YOUTUBE_TEST_OWNED_VIDEO_ID=
+
+# -- X (formerly Twitter) --
+# OAuth2 user access token with scopes: tweet.read, tweet.write, users.read,
```
TheRealAgentK: Please confirm these scopes are all needed. The validator flagged `users.read` and `offline.access` as potentially unused. Also, before approval we need to see the real integration test results with a valid X token, not just the skip-cleanly run without credentials.

TheRealAgentK (commented):
Left two follow-up comments.

TheRealAgentK (changes requested):
Requesting changes based on the inline findings. We need the follow-ups addressed and proper integration test results before approval.

Tram-Nguyen87:
Thanks Kai — update on your comments:

**1. `search_tweets` error handling (x.py)** — Fixed and pushed (`2970941`). It now reads the response body and returns an `ActionError` when X sends an `errors` body, before building the success result — same pattern as the create/bookmark paths. Added a unit test (`test_api_error_returns_action_error`) covering that branch; full unit suite green (69 passed).

**2. Scopes (`users.read`, `offline.access`)** — Confirmed both are needed; the validator can't see them because neither maps to a literal endpoint call:
- `users.read` powers `GET /users/me`, `GET /users/{id}` / `by/username`, and the `author_id` expansions we request on search/timeline/liked/bookmarks (without it those user objects come back empty).
- `offline.access` is the scope that makes X issue a **refresh token** — it's consumed by the platform's token-refresh layer, not by any call in the integration, so it never appears in the code. Dropping it would break connections after the access token expires (~2h). Keeping both.

**3. Real integration test results** — Not done yet, and I want to be upfront: we're currently blocked on generating a valid X token due to an account-access issue on our side (Ninos is sorting it). As soon as that's resolved I'll run the live e2e tests against a real token and post the actual results here before asking for re-review. Holding the PR on that.

.env.example
```diff
@@ -168,3 +168,10 @@
...
+# follows.read, follows.write, like.read, bookmark.read, bookmark.write, offline.access
```
TheRealAgentK: This scope list is missing `media.write`, but `x/config.json` and the README both include it, and `create_tweet` supports media upload. Please add `media.write` here too so the live-test token setup instructions match the integration’s actual required scopes.

TheRealAgentK (commented):
One additional docs follow-up: `x/README.md` still documents the old SDK 1.x-style `result` / `error` fields as normal action output and the removed `python tests/test_x.py` workflow. Since this PR moves failures to `ActionError`, removes those fields from `config.json`, and replaces the old script with pytest unit/integration tests, please update the README so the public docs match the new behavior and test commands.

Tram-Nguyen87:
**README (`x/README.md`)** — brought in line with the SDK 2.0 behavior:
- The "Action Results" section and every action's **Outputs** no longer list the old `result`/`error` fields. It now states that success returns data-only results and that failures (X API `errors` body, non-2xx status, rate limiting, invalid input) raise an `ActionError` surfaced as a failed step.
- The Testing section drops the removed `python tests/test_x.py` workflow and documents the pytest suites instead — `pytest x/tests/` for the default unit run, `-m integration` (and `-m "integration and destructive"`) for the live e2e tests, with the `X_ACCESS_TOKEN` / `X_TEST_TARGET_USER_ID` env vars and the skip-cleanly behavior. Also fixed the stale `cd x-twitter` path and added a version-history entry.

**`.env.example`** — added the missing `media.write` scope to the X token comment so the live-test setup matches `config.json`, the README, and `create_tweet`'s media upload path.

Tram-Nguyen87:
Ran the full live e2e tests locally — 13/16 integration tests pass, 3 skipped (read-only get_me, get_user, get_account_info, search_tweets, get_user_tweets, get_liked_tweets, get_bookmarks + input validation all green; destructive lifecycles covered create→delete tweet, bookmark→remove, and follow→unfollow against a test account, each self-reverting). The 3 skips aren't failures — they chain off a search_tweets result that came back empty on this token's API tier (repost→unrepost, get_tweet by id, post structure). All 69 unit tests also passing.

TheRealAgentK (dismissed):
This is ok now - might need a re-approval after conflict resolution.

TheRealAgentK (approved):

PR #357, opened 2026-06-10
feat(fergus): migrate to autohive-integrations-sdk 2.0.0

Shubhank-Jonnada:
## Summary

- Bumps `autohive-integrations-sdk` from `~=1.0.2` to `~=2.0.0`
- Bumps `config.json` version to `2.0.0`
- Adds `.data` access on all `context.fetch()` call responses
- Replaces `_error()` helper with `ActionError` raises (SDK 2.0 error pattern)
- Fixes custom auth token extraction: `context.auth.get("api_token")` — flat object, no nested credentials envelope
- Cleans `ActionResult` data: removes `result: True/False` bool fields

## Tests

Replaced legacy `tests/context.py` and old integration test file with:
- `tests/conftest.py` — `mock_context` fixture only, no `sys.path` manipulation
- `tests/test_fergus_unit.py` — 17 unit tests covering all 9 actions (`create_job`, `update_job`, `finalise_job`, `get_job`, `list_jobs`, `search_customers`, `get_customer`, `list_sites`, `list_users`)
- `tests/test_fergus_integration.py` — live API tests (skip when `FERGUS_API_TOKEN` not set)

## Validation

- `validate_integration.py fergus` → ✅ 0 errors, 0 warnings
- `check_code.py fergus` → ✅ all checks passed
- `pytest fergus/tests/test_fergus_unit.py` → ✅ 17/17 passed

## Env vars

Added to `.env.example`:
- `FERGUS_API_TOKEN`
- `FERGUS_TEST_JOB_ID`
- `FERGUS_TEST_CUSTOMER_ID`
- `FERGUS_TEST_SITE_ID`


Shubhank-Jonnada:
All unit tests performed and passing. Live integration tests not yet run — API key required.

fergus/tests/test_fergus_integration.py
```diff
@@ -1,54 +1,60 @@
...
+FERGUS_TEST_SITE_ID = os.getenv("FERGUS_TEST_SITE_ID", "")
+
+
+@pytest.fixture
+def live_context():
+    if not FERGUS_API_TOKEN:
+        pytest.skip("FERGUS_API_TOKEN not set")
+    ctx = ExecutionContext.__new__(ExecutionContext)
+    ctx.auth = {"api_token": FERGUS_API_TOKEN}
+    return ctx
```
TheRealAgentK, line 41: Blocker: this bypasses `ExecutionContext.__init__`, so fields like `_session`, config, and logger are missing. The first real `context.fetch()` will fail before reaching Fergus. Please use `ExecutionContext(auth={"api_token": ...})`, or an aiohttp-backed `real_fetch` fixture that returns `FetchResponse`.

fergus/tests/test_fergus_integration.py
```diff
@@ -1,54 +1,60 @@
...
+@pytest.mark.asyncio
+async def test_list_jobs_live(live_context):
+    result = await fergus.execute_action("list_jobs", {"page_size": 5}, live_context)
+    assert result.result.data.get("jobs") is not None
+
+
+@pytest.mark.asyncio
+async def test_list_users_live(live_context):
+    result = await fergus.execute_action("list_users", {"page_size": 5}, live_context)
+    assert result.result.data.get("users") is not None
```
TheRealAgentK, line 55: Should fix: the integration tests assert directly on `result.result.data` without checking `result.type`. If an action returns `ACTION_ERROR` or `VALIDATION_ERROR`, the failure will be noisy and less useful. Please assert `result.type == ResultType.ACTION` first, with the error message included on failure.

fergus/tests/test_fergus_unit.py
```diff
@@ -0,0 +1,181 @@
+from unittest.mock import AsyncMock, MagicMock
+import pytest
+from autohive_integrations_sdk import ResultType
+from fergus.fergus import fergus
+
+pytestmark = pytest.mark.unit
+
+
```
TheRealAgentK, line 8: Should fix: these unit tests use a local `MagicMock(data=...)` helper instead of the SDK response type. They pass, but for an SDK 2 migration the mocks should use `FetchResponse(status=200, headers={}, data=...)` so the tests match the `context.fetch()` contract.

fergus/fergus.py
```diff
@@ -20,48 +19,15 @@ def _auth_headers(api_token: str, include_content_type: bool = False) -> Dict[st
...
-                "status_code": e.status,
-                "response": e.response_data,
-            }
-        )
-    return ActionResult(data={"result": False, "error": str(e), "error_type": type(e).__name__})
-
-
 def _get_token(context: ExecutionContext) -> str:
-    token = context.auth.get("credentials", {}).get("api_token")
+    token = context.auth.get("api_token")
```
TheRealAgentK: Note: I do not agree with the open comment saying this should read from a `credentials` wrapper. Based on the installed SDK 2 behavior and our unit-test pattern, custom auth fields are validated directly against `context.auth`, so this flat `{"api_token": ...}` shape is consistent here.

TheRealAgentK (changes requested):
Requesting changes for the live integration test context issue. Please also tighten the result assertions and unit-test response mocks.

TheRealAgentK:

fergus/tests/test_fergus_integration.py
```diff
@@ -1,54 +1,60 @@
...
+    result = await fergus.execute_action("get_job", {"job_id": int(FERGUS_TEST_JOB_ID)}, live_context)
+    assert result.result.data.get("job") is not None
+
+
+@pytest.mark.asyncio
+async def test_get_customer_live(live_context):
+    if not FERGUS_TEST_CUSTOMER_ID:
+        pytest.skip("FERGUS_TEST_CUSTOMER_ID not set")
+    result = await fergus.execute_action("get_customer", {"customer_id": int(FERGUS_TEST_CUSTOMER_ID)}, live_context)
+    assert result.result.data.get("customer") is not None
```
TheRealAgentK, line 87: Coverage follow-up: the live suite only covers read/list paths. The write lifecycle actions `create_job`, `update_job`, and `finalise_job` are not exercised against Fergus. Please add a guarded destructive lifecycle test, or document why those actions cannot be covered live.

TheRealAgentK (commented):
Added a coverage follow-up.

fergus/tests/test_fergus_integration.py
Shubhank-Jonnada: Fixed in f18b1ca - replaced ExecutionContext.__new__ with the standard make_context + aiohttp side_effect pattern. live_context now routes every call through the real Fergus API when FERGUS_API_TOKEN is set.

fergus/tests/test_fergus_integration.py
Shubhank-Jonnada: Fixed in a3fc6a3 - all integration tests now assert result.type == ResultType.ACTION first, with the error message included on failure, before checking the data fields.

fergus/tests/test_fergus_unit.py
Shubhank-Jonnada: Fixed in 1e3e193 - _fetch_result now returns FetchResponse(status=200, headers={}, data=...) matching the real context.fetch() contract. MagicMock import removed entirely.

fergus/tests/test_fergus_integration.py
Shubhank-Jonnada: Fixed in 5f949dc - added test_job_lifecycle_live marked @pytest.mark.destructive covering create_job, update_job, and finalise_job end-to-end. Finalise is used as cleanup since Fergus exposes no delete endpoint for jobs. Requires FERGUS_TEST_CUSTOMER_ID and FERGUS_TEST_SITE_ID.

Shubhank-Jonnada:
image

TheRealAgentK (approved):

PR #358, opened 2026-06-10
feat(ghost): migrate to autohive-integrations-sdk 2.0.0

Shubhank-Jonnada:
## Summary

- Bumps `autohive-integrations-sdk` from `~=1.0.2` to `~=2.0.0`; keeps `PyJWT>=2.0,<3` for JWT generation
- Bumps `config.json` version to `2.0.0`; cleans output schemas
- Replaces synchronous `requests`-based helpers with async `context.fetch()` calls
- Fixes custom auth: `context.auth.get("api_url/content_api_key/admin_api_key")` — flat object, no nested credentials envelope
- Adds `.data` access on all `context.fetch()` responses
- Replaces `_error()` helper with `ActionError` raises
- Removes `result`/`error`/`error_type` bool fields from `ActionResult` data

## Tests

Replaced legacy `tests/context.py` and `tests/test_ghost.py` with:
- `tests/conftest.py` — `mock_context` fixture with all 3 Ghost auth fields, no `sys.path` manipulation
- `tests/test_ghost_unit.py` — 22 unit tests covering all 15 actions (both content API and admin API paths)
- `tests/test_ghost_integration.py` — live API tests (skip when env vars not set)

## Validation

- `validate_integration.py ghost` → ✅ 0 errors, 0 warnings
- `check_code.py ghost` → ✅ all checks passed (scope warnings are false positives from dynamic dict comprehension patterns)
- `pytest ghost/tests/test_ghost_unit.py` → ✅ 22/22 passed

## Env vars

Added to `.env.example`:
- `GHOST_API_URL`
- `GHOST_CONTENT_API_KEY`
- `GHOST_ADMIN_API_KEY`
- `GHOST_TEST_POST_ID`
- `GHOST_TEST_POST_UPDATED_AT`
- `GHOST_TEST_NEWSLETTER_SLUG`


Shubhank-Jonnada:
All unit tests performed and passing. Live integration tests not yet run — API key required.

ghost/ghost.py
```diff
@@ -1,399 +1,371 @@
...
+            filename = file_path.replace("\\", "/").split("/")[-1]
+            files = {
+                "file": (filename, file_data, mime_type),
+                "purpose": (None, purpose),
+            }
+            resp = await context.fetch(
+                f"{base}/ghost/api/admin/images/upload/",
+                method="POST",
+                headers=headers,
+                files=files,
```
TheRealAgentK: Blocker: `ExecutionContext.fetch()` in SDK 2 does not support `files=`, so `upload_image` will fail before it reaches Ghost. Please either implement this with the supported fetch arguments or keep a direct HTTP client for this multipart upload path.

ghost/tests/test_ghost_integration.py
```diff
@@ -0,0 +1,72 @@
...
+GHOST_TEST_POST_ID = os.getenv("GHOST_TEST_POST_ID", "")
+GHOST_TEST_POST_UPDATED_AT = os.getenv("GHOST_TEST_POST_UPDATED_AT", "")
+GHOST_TEST_NEWSLETTER_SLUG = os.getenv("GHOST_TEST_NEWSLETTER_SLUG", "")
+
+
+@pytest.fixture
+def live_context():
+    if not GHOST_API_URL or not GHOST_CONTENT_API_KEY or not GHOST_ADMIN_API_KEY:
+        pytest.skip("GHOST_API_URL, GHOST_CONTENT_API_KEY, and GHOST_ADMIN_API_KEY must be set")
+    ctx = ExecutionContext.__new__(ExecutionContext)
```
TheRealAgentK: Blocker: this bypasses `ExecutionContext.__init__`, so the context never gets a real `fetch` implementation. The live tests will not exercise SDK 2 request handling. Please create the context through the normal SDK path and include a passing run with real Ghost credentials.

ghost/tests/test_ghost_integration.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@pytest.mark.asyncio
+async def test_get_tiers_live(live_context):
+    result = await ghost.execute_action("get_tiers", {}, live_context)
+    assert result.result.data.get("tiers") is not None
+
+
+@pytest.mark.asyncio
+async def test_create_draft_post_live(live_context):
```
TheRealAgentK: Blocker: this creates a real draft post but is not marked `destructive`, so it runs with the read-only integration tests. Please mark it destructive, or add cleanup that removes the created post.

ghost/tests/test_ghost_unit.py
```diff
@@ -0,0 +1,213 @@
+from unittest.mock import AsyncMock, MagicMock
+import pytest
+from autohive_integrations_sdk import ResultType
+from ghost.ghost import ghost
+
+pytestmark = pytest.mark.unit
+
+
+def _fetch_result(data):
```
TheRealAgentK, line 9: Should fix: these unit tests should use the SDK `FetchResponse` shape instead of a loose `MagicMock`. That catches `.data` usage and response-shape mistakes closer to how SDK 2 behaves.

ghost/tests/conftest.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@pytest.fixture
+def mock_context():
+    ctx = MagicMock(name="ExecutionContext")
+    ctx.fetch = AsyncMock(name="fetch")
+    ctx.auth = {
+        "api_url": "https://demo.ghost.io",
+        "content_api_key": "test_content_key",  # nosec B105
+        "admin_api_key": "6747844835dbba000136a9b3:f9e9e900c9e04cde3cc42d387c8757490a0d81db5a44e49e46af0dbd0ef64ec1",  # nosec B105
```
TheRealAgentK: Should fix: this looks like a real Ghost admin key format and secret. Even if it is only test data, please replace it with an obviously fake value like `test_key_id:aabbcc...` so it does not trip reviews or secret-scanning follow-ups.

ghost/ghost.py
```diff
@@ -1,399 +1,371 @@
...
 ghost = Integration.load()
 
 
-# ---- Helpers ----
-
-
 def _get_base_url(context: ExecutionContext) -> str:
-    credentials = context.auth.get("credentials", {})
-    url = credentials.get("api_url", "")
+    url = context.auth.get("api_url", "")
```
TheRealAgentK, line 19: Note: this flat custom-auth access is the right SDK 2 shape for Ghost. Please keep it direct on `context.auth` and avoid adding wrappers that recreate the old nested `credentials` shape.

TheRealAgentK (commented):
Left blockers and follow-ups for the Ghost SDK 2 migration. The live integration tests need to pass with real Ghost credentials before approval.

TheRealAgentK (changes requested):
Requesting changes based on the blockers and follow-ups. The live integration tests need to pass with real Ghost credentials before approval.

ghost/tests/test_ghost_integration.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@pytest.mark.asyncio
+async def test_get_settings_live(live_context):
+    result = await ghost.execute_action("get_settings", {}, live_context)
+    assert result.result.data.get("settings") is not None
+
+
+@pytest.mark.asyncio
+async def test_get_tiers_live(live_context):
```
TheRealAgentK, line 99: Coverage follow-up: the live suite only covers a small read subset plus create post. It does not exercise `get_post`, `get_page`, `update_post`, `upload_image`, `create_page`, member actions, or `send_newsletter`. Please add safe read coverage where possible and marked destructive coverage for admin write paths that we expect to support.

TheRealAgentK (commented):
Added a coverage follow-up.

TheRealAgentK:
Follow-up from re-review: `ghost/tests/test_ghost_integration.py` reads `GHOST_TEST_MEMBER_ID`, but the Ghost section in root `.env.example` does not document it. Per the integration-test checklist, every env var read by integration tests should be represented there. Could you either add a blank/commented `GHOST_TEST_MEMBER_ID=` entry, or remove the env read if it is dead/unused?

ghost/ghost.py
```diff
@@ -1,399 +1,382 @@
...
+                    f"{purpose}\r\n"
+                    f"--{boundary}--\r\n"
+                ).encode()
+            )
+            headers["Content-Type"] = f"multipart/form-data; boundary={boundary}"
+            resp = await context.fetch(
+                f"{base}/ghost/api/admin/images/upload/",
+                method="POST",
+                headers=headers,
+                body=body,
```
TheRealAgentK: 🚫 **Blocker:** this still passes the multipart payload via `body=`, but SDK 2 `ExecutionContext.fetch()` does not support a `body` keyword. It supports request bodies through `data=` or `json=`.

This is the same class of issue as the earlier Ghost `upload_image` upload-path feedback: the upload action needs to use only fetch arguments that the real SDK accepts. Right now the live test fixture accepts `body` and maps it to `data`, so the integration test can pass while production SDK execution would fail before the request reaches Ghost.

Please change this call to use `data=body` (keeping the multipart `Content-Type` header), and update the live fixture to accept/use `data=` rather than `body=` so tests mirror the SDK 2 contract and catch this regression.

Shubhank-Jonnada:
image

TheRealAgentK (approved):

PR #359, opened 2026-06-10
feat(circle): migrate to autohive-integrations-sdk 2.0.0

Shubhank-Jonnada:
## Summary

- Bumps `autohive-integrations-sdk` from `~=1.0.2` to `~=2.0.0`; keeps `mistune>=3.0.0` for markdown→TipTap conversion
- Bumps `config.json` version to `2.0.0`; removes `result` bool from output schemas
- Adds `.data` access on all `context.fetch()` responses (SDK 1.x returned raw dicts)
- Fixes custom auth: `context.auth.get("api_token")` — flat object, no credentials envelope
- Replaces `handle_api_response()` / error `ActionResult` returns with `ActionError` returns
- `_check_api_error()` raises `ValueError` caught by the outer try/except → `return ActionError(message=...)`
- TipTap/ProseMirror markdown renderer kept unchanged (pure application logic)

## Tests

Replaced legacy `tests/context.py` and `tests/test_circle.py` with:
- `tests/conftest.py` — `mock_context` fixture, no `sys.path` manipulation
- `tests/test_circle_unit.py` — 27 unit tests covering all 23 actions
- `tests/test_circle_integration.py` — live API tests (skip when env vars not set)

## Validation

- `validate_integration.py circle` → ✅ 0 errors, 0 warnings
- `check_code.py circle` → ✅ all checks passed (scope warnings are false positives from `build_search_params`)
- `pytest circle/tests/test_circle_unit.py` → ✅ 27/27 passed

## Env vars

Added to `.env.example`:
- `CIRCLE_API_TOKEN`
- `CIRCLE_TEST_POST_ID`
- `CIRCLE_TEST_SPACE_ID`
- `CIRCLE_TEST_MEMBER_ID`


Shubhank-Jonnada:
All unit tests and live integration tests performed. 13/14 read-only actions passed against the live API (1 skipped — needs `CIRCLE_TEST_MEMBER_EMAIL`).

Shubhank-Jonnada:
Regarding the bot comment on `build_auth_headers`: the flat `context.auth.get("api_token")` lookup is intentional and correct for SDK 2.0. The nested `context.auth.get("credentials", {}).get("api_token")` was the SDK 1.x pattern — this migration specifically removes that wrapper. In SDK 2.0, `context.auth` is a flat dict matching the fields defined in `config.auth.fields` directly.

.env.example
```diff
@@ -168,3 +168,10 @@
 # YOUTUBE_TEST_THUMBNAIL_PATH=
 # Video ID owned by the authenticated user (required by moderate_comment, which only works on your own channel)
 # YOUTUBE_TEST_OWNED_VIDEO_ID=
+
+# -- Circle --
+# CIRCLE_API_TOKEN=
+# Optional — target specific resources for integration tests
+# CIRCLE_TEST_POST_ID=
+# CIRCLE_TEST_SPACE_ID=
+# CIRCLE_TEST_MEMBER_ID=
```
TheRealAgentK: Should fix: `circle/tests/test_circle_integration.py` reads `CIRCLE_TEST_MEMBER_EMAIL`, but this Circle env section does not document it. Please add `# CIRCLE_TEST_MEMBER_EMAIL=` here.

circle/tests/test_circle_integration.py
```diff
@@ -0,0 +1,227 @@
...
+                params=params,
+                json=json,
+                data=body,
+                headers=dict(headers or {}),
+            ) as resp:
+                try:
+                    data = await resp.json(content_type=None)
+                except Exception:
+                    data = await resp.text()
+                return FetchResponse(status=resp.status, headers=dict(resp.headers), data=data)
```
TheRealAgentK, line 70: Should fix: this returns `FetchResponse` for non-2xx responses, but SDK `ExecutionContext.fetch()` raises on non-2xx. The source usually converts API error bodies into `ActionError`, but this fixture is still less production-realistic than it should be. Please mirror SDK behavior: raise `RateLimitError` for 429 and `HTTPError` for other non-2xx statuses.

TheRealAgentK (changes requested):
Requesting changes for the integration test env docs and `real_fetch` behavior.

TheRealAgentK:

circle/tests/test_circle_integration.py
```diff
@@ -0,0 +1,227 @@
...
+    assert result.type != ResultType.ACTION_ERROR, result.result.message
+    assert isinstance(result.result.data.get("space_groups"), list)
+
+
+@skip_if_no_token
+@pytest.mark.asyncio
+async def test_list_access_groups(live_context):
+    result = await circle.execute_action("list_access_groups", {}, live_context)
+    assert result.type != ResultType.ACTION_ERROR, result.result.message
+    assert isinstance(result.result.data.get("access_groups"), list)
```
TheRealAgentK, line 240: Coverage follow-up: live coverage is read-heavy. The mutation actions are not exercised here: `create_post`, `update_post`, `create_comment`, member tag add/remove, space group add/remove, and access group add/remove. Please add marked destructive tests for these, or document why each cannot be safely covered live.

TheRealAgentK (commented):
Added a coverage follow-up.

.env.example
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/c7065a7052867793471b4b0820892974703e52b8 - added `CIRCLE_TEST_MEMBER_EMAIL=` to the Circle section of `.env.example`.

circle/tests/test_circle_integration.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/968e0d68bcb9b3ab200db285764f32c6789ebe84 - `real_fetch` now raises `RateLimitError` for 429 and `HTTPError` for all other non-2xx, matching SDK behavior.

circle/tests/test_circle_integration.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/219060308719799d84da510b1fd5f90cfe66e5e4 - added `@pytest.mark.destructive` tests for all mutation actions: `create_post` + `update_post` lifecycle, `create_comment`, `add_member_tags`/`remove_member_tags`, `add_member_to_space_groups`/`remove_member_from_space_groups`, and `add_member_to_access_groups`/`remove_member_from_access_groups`.

circle/circle.py
```diff
@@ -288,50 +263,31 @@ def text_to_tiptap_body(text: str) -> Dict[str, Any]:
...
+    return md(text)
 
 
 # ---- Utility Functions ----
 
 
 def build_auth_headers(context: ExecutionContext) -> Dict[str, str]:
-    """Build authorization headers for Circle API"""
-    api_token = context.auth.get("credentials", {}).get("api_token")
+    api_token = context.auth.get("api_token")
```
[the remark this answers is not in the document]
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/6a652d6 - `build_auth_headers` now reads from `context.auth["credentials"]["api_token"]` first (production shape) and falls back to the flat `context.auth["api_token"]` so existing unit-test mocks continue to work.

Shubhank-Jonnada:
@TheRealAgentK 
image

circle/circle.py
```diff
@@ -342,133 +298,79 @@ class SearchPostsAction(ActionHandler):
...
-            if "is_pinned" in inputs:
-                update_data["is_pinned"] = inputs["is_pinned"]
-
-            # Make API call
-            response = await context.fetch(
+            update_data: Dict[str, Any] = {}
+            if inputs.get("name") is not None:
+                update_data["name"] = inputs.get("name")
+            if inputs.get("body") is not None:
+                update_data["body"] = inputs.get("body")
```
TheRealAgentK: Thanks for expanding the Circle SDK v2 migration here. One thing I’d like to tighten before approval: `update_post` still sends body updates as a raw `body` field, while `create_post` converts post content into Circle’s TipTap payload via `tiptap_body`. Since `update_post` exposes `body` in the input schema, we should probably use the same payload shape there too (e.g. `tiptap_body: {"body": text_to_tiptap_body(...)}`), and add coverage for updating the body. The current destructive lifecycle test only updates the title, so it won’t catch this path.

circle/tests/test_circle_integration.py
```diff
@@ -0,0 +1,391 @@
...
+async def test_create_comment(live_context):
+    """create_comment on an existing or freshly created post."""
+    post_id = CIRCLE_TEST_POST_ID
+    if not post_id:
+        space_id = CIRCLE_TEST_SPACE_ID
+        if not space_id:
+            pytest.skip("CIRCLE_TEST_POST_ID or CIRCLE_TEST_SPACE_ID required for create_comment")
+        create_result = await circle.execute_action(
+            "create_post",
+            {"space_id": space_id, "name": "Autohive comment test post", "body": "test"},
```
TheRealAgentK: Same destructive-test safety point here: when no `CIRCLE_TEST_POST_ID` is provided, this fallback creates a new post for comment testing and can leave visible test content behind. Please make this safe for repeated live runs — ideally draft/test-only content plus cleanup if Circle supports it, or explicit documentation that destructive tests must target a dedicated test space/community.

circle/tests/test_circle_integration.py
```diff
@@ -0,0 +1,391 @@
...
+@pytest.mark.asyncio
+async def test_create_update_post(live_context):
+    """create_post -> update_post lifecycle."""
+    space_id = CIRCLE_TEST_SPACE_ID
+    if not space_id:
+        pytest.skip("CIRCLE_TEST_SPACE_ID not set — required for create_post")
+
+    create_result = await circle.execute_action(
+        "create_post",
+        {"space_id": space_id, "name": "Autohive integration test post", "body": "test body"},
```
TheRealAgentK: Nice to see the destructive coverage added. One safety concern: this creates a post without specifying `status`, so `create_post` defaults to `published`, and there doesn’t appear to be cleanup afterwards. That means a live destructive test run can leave visible test content in the Circle community. If Circle doesn’t expose a delete action here, could we at least create draft/test-only content where supported, require a dedicated test space, and document that limitation?

circle/config.json
```diff
@@ -539,18 +529,17 @@
...
-        "required": ["user_email", "member_tag_ids"]
+        "required": [
+          "user_email",
+          "member_tag_ids"
+        ]
       },
       "output_schema": {
         "type": "object",
         "properties": {
           "member": {
```
TheRealAgentK, line 552: Small schema sync issue: several mutation actions now return count fields that aren’t reflected in the output schemas. For example, `add_member_tags` returns `tags_added`, but this schema only documents `member`; `remove_member_tags` returns `tags_removed` while its schema still documents `member`. Same pattern applies to the space/access group add/remove actions. Could you update these output schemas so they match the actual successful action results?

TheRealAgentK (commented):
Thanks for the follow-up work here — the Circle integration test situation is much better now: read-only/destructive commands are documented, tests are marked correctly, non-2xx behavior now mirrors SDK behavior, env vars are documented, and mutation coverage was added.

I left a few comments on the remaining items I’d like tightened before treating the live coverage as fully credible: body updates should use the same TipTap payload shape as creates and be covered by tests, the mutation output schemas should match the new count fields, and destructive tests should avoid leaving visible/published test content behind.

Shubhank-Jonnada:
Addressed all reviewer feedback — four commits on top of the existing work:

- **`a35ae3a`** — `update_post` now sends `tiptap_body: {"body": text_to_tiptap_body(...)}` matching the `create_post` payload shape; destructive lifecycle test updated to also exercise the body update path
- **`69b8c05`** — Destructive tests create posts as `draft` and DELETE them afterwards via the admin API; no visible test content left in the community on repeated runs. `test_create_comment` fallback also creates draft + cleans up
- **`115598d`** — Output schemas synced with actual return values: `add_member_tags` → adds `tags_added`, `remove_member_tags` → `tags_removed` (removes stale `member`), same pattern for space/access group add/remove actions
- **`3a20bbb`** — All config-code sync warnings resolved by inlining `inputs.get()` calls explicitly; `page` param added to `search_posts`/`search_spaces`/`search_events` input schemas where it was missing

27/27 unit tests passing, `check_code` fully clean (zero warnings). Ready for live integration test run with `CIRCLE_API_TOKEN` + `CIRCLE_TEST_SPACE_ID`.

circle/circle.py
Shubhank-Jonnada: Fixed in `a35ae3a` — `update_post` now sends `tiptap_body: {"body": text_to_tiptap_body(...)}` matching the `create_post` shape. Destructive lifecycle test updated to pass `body` so this path is exercised on every live run.

circle/tests/test_circle_integration.py
Shubhank-Jonnada: Fixed in `69b8c05` — when `CIRCLE_TEST_POST_ID` is not set, the fallback now creates a draft post (`status: draft`) and deletes it after the comment is created. No visible test content is left in the community.

circle/tests/test_circle_integration.py
Shubhank-Jonnada: Fixed in `69b8c05` — `test_create_update_post` now creates with `status: draft` and DELETE cleans up the post at the end. The doc comment explains this requires a dedicated test space.

circle/config.json
Shubhank-Jonnada: Fixed in `115598d` — output schemas updated to match actual return values: `add_member_tags` → `{member, tags_added}`, `remove_member_tags` → `{tags_removed}`, `add_member_to_space_groups` → `{member, groups_added}`, `remove_member_from_space_groups` → `{groups_removed}`, same for access group actions.

Shubhank-Jonnada:
Live test results:
tested read-only actions only, as this uses our main community platform key.

TheRealAgentK (approved):

PR #360, opened 2026-06-10
chore(aws): migrate to autohive-integrations-sdk 2.0.0

Shubhank-Jonnada:
## Summary
- Bump autohive-integrations-sdk to ~=2.0.0 in requirements.txt
- Fix auth: flat context.auth.get() instead of nested credentials dict
- Update helpers.py: error_result() now returns ActionError dataclass; success_result() drops result:True and adds cost_usd=0.0
- Bump config.json version to 2.0.0; remove result/error/error_code fields from all 20 output schemas
- Replace legacy tests/context.py + tests/test_aws.py with conftest.py + test_aws_unit.py + test_aws_integration.py

## Test plan
- [ ] 25 unit tests, all 20 actions covered: pytest aws/tests/test_aws_unit.py
- [ ] validate_integration.py autohive-integrations/aws - 0 errors, 0 warnings
- [ ] check_code.py autohive-integrations/aws - all checks passed
- [ ] Live integration tests require AWS credentials (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY)


Shubhank-Jonnada:
All unit tests performed and passing. Live integration tests not yet run — AWS credentials required.

aws/tests/test_aws_integration.py
```diff
@@ -0,0 +1,91 @@
...
+
+AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID", "")
+AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY", "")
+AWS_REGION = os.getenv("AWS_REGION", "us-east-1")
+AWS_DETECTOR_ID = os.getenv("AWS_DETECTOR_ID", "")
+AWS_TRAIL_NAME = os.getenv("AWS_TRAIL_NAME", "")
+AWS_LOG_GROUP = os.getenv("AWS_LOG_GROUP", "")
+AWS_LOG_STREAM = os.getenv("AWS_LOG_STREAM", "")
+
+pytestmark = pytest.mark.skipif(
```
TheRealAgentK: Blocker: this file is missing `pytest.mark.integration`. Under the documented `-m integration` command, all AWS integration tests are deselected. Please use:

```python
pytestmark = [
    pytest.mark.integration,
    pytest.mark.skipif(
        not AWS_ACCESS_KEY_ID or not AWS_SECRET_ACCESS_KEY,
        reason="AWS credentials not set in environment",
    ),
]
```

aws/tests/test_aws_integration.py
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    class _Ctx:
+        auth = ctx_auth
+
+    return _Ctx()
+
+
+@pytest.mark.asyncio
+async def test_get_findings_live(live_context):
```
TheRealAgentK: Blocker: `ResultType.SUCCESS` does not exist in SDK 2.0. Successful action results are `ResultType.ACTION`, so these assertions need to use that.

TheRealAgentK: Blocker: every live test allows `ACTION_ERROR` as a passing outcome. Invalid credentials or missing permissions can still pass the suite. For at least the baseline read-only tests, assert `result.type == ResultType.ACTION`. If a service may not be enabled, skip based on a clear precondition rather than accepting `ACTION_ERROR`.

TheRealAgentK (changes requested):
Requesting changes for the AWS integration test blockers.

TheRealAgentK:

aws/tests/test_aws_integration.py
```diff
@@ -0,0 +1,91 @@
...
+    if result.type == ResultType.SUCCESS:
+        assert "events" in result.result.data
+
+
+@pytest.mark.asyncio
+async def test_describe_trails_live(live_context):
+    result = await aws.execute_action("describe_trails", {}, live_context)
+    assert result.type in (ResultType.SUCCESS, ResultType.ACTION_ERROR)
+    if result.type == ResultType.SUCCESS:
+        assert "trails" in result.result.data
```
TheRealAgentK: Coverage follow-up: this live suite covers only 7 of the 20 AWS actions. Missing live coverage includes Security Hub details/update/insights, GuardDuty finding detail/archive, CloudWatch metric data/alarm history/set state, CloudWatch Logs filter/get events, and CloudTrail status/selectors. It is fine to gate service-specific tests behind env vars or clear preconditions, but the migration should not leave these paths completely untested live.

TheRealAgentK (commented):
Added a coverage follow-up.

aws/helpers.py
```diff
@@ -5,20 +5,19 @@
...
-from autohive_integrations_sdk import ActionResult, ExecutionContext
+from autohive_integrations_sdk import ActionError, ActionResult, ExecutionContext
 
 
 def create_boto3_client(context: ExecutionContext, service_name: str):
-    credentials = context.auth.get("credentials", {})
-    access_key = credentials.get("aws_access_key_id")
-    secret_key = credentials.get("aws_secret_access_key")
+    access_key = context.auth.get("aws_access_key_id")
+    secret_key = context.auth.get("aws_secret_access_key")
```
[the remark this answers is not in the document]
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/701c068 - `create_boto3_client` now reads from `context.auth['credentials']` first (production shape), falling back to the flat dict so unit-test mocks continue to work.

aws/tests/test_aws_integration.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/4e0ef60 - `pytestmark` is now a list combining `pytest.mark.integration` with the skipif condition.

aws/tests/test_aws_integration.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/4e0ef60 - all live-path assertions use `assert result.type == ResultType.ACTION`.

Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/4e0ef60 - `ACTION_ERROR` removed as a passing outcome; every test now asserts `== ResultType.ACTION`. Service-specific tests that need env vars skip cleanly rather than accepting errors.

aws/tests/test_aws_integration.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/4e0ef60 - all 20 actions now have live coverage. Added tests for `get_finding_details`, `get_insights`, `update_finding_workflow`, `list_guardduty_findings`, `get_guardduty_finding_details`, `archive_findings`, `get_metric_data`, `get_alarm_history`, `set_alarm_state`, `filter_log_events`, `get_log_events`, `get_trail_status`, `get_event_selectors`. Destructive tests are marked `@pytest.mark.destructive`; service-specific tests skip when required env vars are absent.

Shubhank-Jonnada:
@TheRealAgentK 

## Live tests:

image

aws/tests/test_aws_integration.py
```diff
@@ -0,0 +1,324 @@
...
+    assert "trails" in result.result.data
+
+
+@pytest.mark.asyncio
+async def test_get_trail_status(live_context):
+    if not AWS_TRAIL_NAME:
+        pytest.skip("AWS_TRAIL_NAME required for get_trail_status")
+    result = await aws.execute_action("get_trail_status", {"trail_name": AWS_TRAIL_NAME}, live_context)
+    assert result.type == ResultType.ACTION, result.result.message
+    assert "is_logging" in result.result.data
```
TheRealAgentK: Thanks for expanding the AWS live coverage to all actions — that’s a big improvement. One small mismatch here: `get_trail_status` returns `{"trail_status": ...}` (and the config schema documents `trail_status`), so this assertion will fail when `AWS_TRAIL_NAME` is provided even if the action itself succeeds. Could you assert `"trail_status" in result.result.data` instead, and optionally inspect fields inside that object?

aws/helpers.py
```diff
@@ -1,24 +1,27 @@
...
 def create_boto3_client(context: ExecutionContext, service_name: str):
-    credentials = context.auth.get("credentials", {})
-    access_key = credentials.get("aws_access_key_id")
-    secret_key = credentials.get("aws_secret_access_key")
+    creds = context.auth.get("credentials") or context.auth
+    access_key = creds.get("aws_access_key_id")
+    secret_key = creds.get("aws_secret_access_key")
     if not access_key or not secret_key:
         raise ValueError("AWS credentials are missing: aws_access_key_id and aws_secret_access_key are required")
+    session_token = creds.get("aws_session_token") or os.environ.get("AWS_SESSION_TOKEN")
```
TheRealAgentK: Could we avoid falling back to the process environment here? Pulling `AWS_SESSION_TOKEN` from ambient env can mix a user’s configured access key/secret with an unrelated runtime token. It would be cleaner and safer to make `aws_session_token` an optional auth field in `config.json` and only pass it to boto3 when it comes from `context.auth`.

.env.example
```diff
@@ -168,3 +168,30 @@
...
+# AWS_DETECTOR_ID=
+# Optional — needed for CloudTrail tests
+# AWS_TRAIL_NAME=
+# Optional — needed for CloudWatch Logs tests
+# AWS_LOG_GROUP=
+# AWS_LOG_STREAM=
+# Optional — needed for CloudWatch alarm tests
+# AWS_ALARM_NAME=
+# Optional — needed for Security Hub finding tests
+# AWS_FINDING_ARN=
```
TheRealAgentK: The integration tests read `AWS_SESSION_TOKEN`, but this env var isn’t documented in `.env.example`. If temporary credentials are supported in the live tests, please add a blank/commented `AWS_SESSION_TOKEN=` entry here as well.

TheRealAgentK (commented):
Thanks for the AWS follow-up work — this is much closer now. The integration test suite has been expanded to cover all 20 actions, destructive tests are marked, and the live-path assertions now fail on non-action results.

I left a few small comments before I’d call this fully clean: one CloudTrail live-test assertion checks the wrong output key, `AWS_SESSION_TOKEN` should be documented if tests support it, and session tokens should ideally come from `context.auth` rather than ambient process env.

aws/tests/test_aws_integration.py
Shubhank-Jonnada: Fixed in `41ea8a3` -assertion now checks `"trail_status" in result.result.data` and `"IsLogging" in result.result.data["trail_status"]` to match the action's actual output shape.

aws/helpers.py
Shubhank-Jonnada: Fixed in `38b8f48` - removed the `os.environ.get("AWS_SESSION_TOKEN")` fallback. Session token now comes exclusively from `context.auth.get("aws_session_token")`, and the `import os` was removed from helpers.py since it's no longer needed.

.env.example
Shubhank-Jonnada: Fixed in `282dc39` - added `# AWS_SESSION_TOKEN=` with an STS temporary credentials note to the AWS section of `.env.example`.

TheRealAgentK:
Follow-up from re-review: `aws/helpers.py` now reads `aws_session_token` from `context.auth`, and `.env.example` documents `AWS_SESSION_TOKEN` for live tests, but `aws/config.json` does not expose `aws_session_token` as an optional custom-auth field. If temporary AWS credentials are intended to be supported in production, please add `aws_session_token` to the custom auth schema as an optional password/text field. Otherwise the integration tests can inject it, but connected users cannot configure it through the integration auth UI.

Shubhank-Jonnada:
live test results:
Tested again; all read-only actions passed. Some actions require additional subscriptions to test so skipped them

TheRealAgentK (approved):

PR #361, opened 2026-06-10
chore(grammarly): migrate to autohive-integrations-sdk 2.0.0

Shubhank-Jonnada:
## Summary
- Bump autohive-integrations-sdk to ~=2.0.0; drop aiohttp dependency
- Fix auth: flat context.auth.get() instead of nested credentials dict
- Replace aiohttp with context.fetch() for all API calls (token, API requests, file upload)
- Error returns use ActionError instead of ActionResult with result:False
- Bump config.json version to 2.0.0; remove result/error fields from all output schemas
- Replace context.py + test_grammarly.py with conftest.py + unit + integration tests

## Test plan
- [ ] 13 unit tests, all 7 actions covered: pytest grammarly/tests/test_grammarly_unit.py
- [ ] validate_integration.py autohive-integrations/grammarly - 0 errors, 0 warnings
- [ ] check_code.py autohive-integrations/grammarly - all checks passed
- [ ] Live tests require Grammarly OAuth2 credentials (GRAMMARLY_CLIENT_ID, GRAMMARLY_CLIENT_SECRET)


Shubhank-Jonnada:
All unit tests performed and passing. Live integration tests not yet run — Grammarly OAuth credentials required.

grammarly/grammarly.py
```diff
@@ -1,68 +1,50 @@
...
-                token_data = await resp.json()
-                return token_data.get("access_token")
-            else:
-                error_text = await resp.text()
-                raise Exception(f"Failed to obtain access token: HTTP {resp.status}: {error_text}")
+    resp = await context.fetch(
+        GRAMMARLY_TOKEN_URL,
+        method="POST",
+        headers={"Content-Type": "application/x-www-form-urlencoded"},
+        body=form_body,
```
TheRealAgentK: Blocker: SDK 2 `ExecutionContext.fetch()` supports `data=` and `json=`, not `body=`. This token request will fail before it reaches Grammarly. Please send the form payload with `data=...` and `content_type="application/x-www-form-urlencoded"`.

grammarly/grammarly.py
```diff
@@ -71,93 +53,44 @@ async def api_request(
...
-                raise Exception(f"File upload failed: HTTP {resp.status}: {error_text}")
-
-
-# ---- Writing Score API Actions ----
+async def upload_file(context: ExecutionContext, upload_url: str, file_content: str) -> bool:
+    await context.fetch(
+        upload_url,
+        method="PUT",
+        headers={"Content-Type": "text/plain"},
+        body=file_content,
```
TheRealAgentK: Blocker: same issue here. `body=` is not supported by SDK 2 `fetch()`, so uploads will fail. Please use `data=file_content` with `content_type="text/plain"`.

grammarly/tests/test_grammarly_integration.py
```diff
@@ -0,0 +1,54 @@
+import os
+import pytest
+from autohive_integrations_sdk import ResultType
+from grammarly.grammarly import grammarly
+
+GRAMMARLY_CLIENT_ID = os.getenv("GRAMMARLY_CLIENT_ID", "")
+GRAMMARLY_CLIENT_SECRET = os.getenv("GRAMMARLY_CLIENT_SECRET", "")
+
+pytestmark = pytest.mark.skipif(
```
TheRealAgentK: Blocker: this file is missing `pytest.mark.integration`, so the documented `-m integration` command deselects every Grammarly integration test. Please include the integration marker in `pytestmark` alongside this skip condition.

grammarly/tests/test_grammarly_integration.py
```diff
@@ -0,0 +1,54 @@
...
+
+@pytest.fixture
+def live_context():
+    class _Ctx:
+        auth = {
+            "client_id": GRAMMARLY_CLIENT_ID,
+            "client_secret": GRAMMARLY_CLIENT_SECRET,
+        }
+
+    return _Ctx()
```
TheRealAgentK: Blocker: this live context only has `auth` and no `fetch` method. Every action calls `context.fetch()`, so these tests cannot call Grammarly. Please use a real `ExecutionContext(auth=...)` or an aiohttp-backed `real_fetch` fixture.

grammarly/tests/test_grammarly_integration.py
```diff
@@ -0,0 +1,54 @@
...
+            "filename": "test.txt",
+            "file_content": (
+                "The quick brown fox jumps over the lazy dog. "
+                "This is a sample text for writing quality analysis. "
+                "It contains multiple sentences to meet the minimum word count requirement."
+            ),
+        },
+        live_context,
+    )
+    assert result.type in (ResultType.SUCCESS, ResultType.ACTION_ERROR)
```
TheRealAgentK: Blocker: `ResultType.SUCCESS` does not exist in SDK 2. Successful actions are `ResultType.ACTION`. This also allows `ACTION_ERROR` to pass, so invalid credentials or broken API calls can look green. Please assert `result.type == ResultType.ACTION` for successful live-path tests.

grammarly/tests/test_grammarly_unit.py
```diff
@@ -0,0 +1,212 @@
+import pytest
+from unittest.mock import AsyncMock, MagicMock
+from autohive_integrations_sdk import ResultType
+from grammarly.grammarly import grammarly
+
+pytestmark = pytest.mark.unit
+
+TOKEN_RESPONSE = MagicMock(data={"access_token": "test_token"})  # nosec B105
```
TheRealAgentK: Should fix: these unit tests use `MagicMock(data=...)` instead of the SDK response type. Prefer `FetchResponse(status=200, headers={}, data=...)` so the tests match the SDK 2 `fetch()` contract.

TheRealAgentK (changes requested):
Requesting changes for the Grammarly SDK 2 blockers.

TheRealAgentK:

grammarly/tests/test_grammarly_integration.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+@pytest.mark.asyncio
+async def test_get_user_analytics_live(live_context):
+    result = await grammarly.execute_action(
+        "get_user_analytics",
+        {"date_from": "2024-01-01", "date_to": "2024-01-31"},
+        live_context,
+    )
+    assert result.type in (ResultType.SUCCESS, ResultType.ACTION_ERROR)
```
TheRealAgentK: Coverage follow-up: the live suite only covers writing-score submit and analytics. It does not cover the writing-score result path, AI detection submit/results, or plagiarism submit/results. Please add live coverage for those flows, even if result checks are gated behind returned request IDs or polling preconditions.

TheRealAgentK (commented):
Added a coverage follow-up.

grammarly/grammarly.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/6e465c2 - token request now uses `data=form_body, content_type="application/x-www-form-urlencoded"` matching the SDK 2 fetch() signature.

grammarly/grammarly.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/6e465c2 - upload now uses `data=file_content, content_type="text/plain"`.

grammarly/tests/test_grammarly_integration.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/293ea3f - `pytestmark` is now a list combining `pytest.mark.integration` with the skipif condition.

grammarly/tests/test_grammarly_integration.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/293ea3f - `live_context` now uses `make_context` with an aiohttp-backed `real_fetch` that raises `HTTPError`/`RateLimitError` for non-2xx, matching the SDK 2 contract.

grammarly/tests/test_grammarly_integration.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/293ea3f - all live-path assertions now use `assert result.type == ResultType.ACTION`.

grammarly/tests/test_grammarly_unit.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/b4163dd - all mock responses replaced with `FetchResponse(status=200, headers={}, data=...)`.

grammarly/tests/test_grammarly_integration.py
Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/commit/293ea3f - added tests for `get_writing_score_results` (polls after submit), `analyze_ai_detection`, `get_ai_detection_results`, `analyze_plagiarism_detection`, and `get_plagiarism_detection_results`. All 7 actions are now covered.

Shubhank-Jonnada:
Live integration tests passed - 7/7, 19-06-2026.

  One fix applied during testing: `test_get_user_analytics` was using `date_from: 2024-01-01` which the Grammarly API rejects with  `date_from_can_not_exceed_one_year`. Updated to a date range within the last year.

  All actions verified against the live API:
  - analyze_writing_score
  - get_writing_score_results
  - get_user_analytics
  - analyze_ai_detection
  - get_ai_detection_results
  - analyze_plagiarism_detection
  - get_plagiarism_detection_results

grammarly/grammarly.py
```diff
@@ -71,93 +54,44 @@ async def api_request(
...
-                return True
-            else:
-                error_text = await resp.text()
-                raise Exception(f"File upload failed: HTTP {resp.status}: {error_text}")
-
-
-# ---- Writing Score API Actions ----
+async def upload_file(context: ExecutionContext, upload_url: str, file_content: str) -> bool:
+    await context.fetch(
+        upload_url,
```
TheRealAgentK: Blocker: this upload path now passes the pre-signed URL to `context.fetch()` as a plain string, which loses the old `yarl.URL(upload_url, encoded=True)` safeguard. These upload URLs are S3-style signed URLs, and aiohttp can canonicalize encoded query parameters (for example `%2F` in `X-Amz-Credential`) when it builds the request URL, invalidating the signature. Please preserve the encoded URL here, e.g. by passing `URL(upload_url, encoded=True)` through to `context.fetch()` if supported, or by keeping a dedicated aiohttp upload helper that uses `encoded=True`. A small unit test around the upload URL/call shape would help prevent this from regressing again.

grammarly/tests/test_grammarly_integration.py
```diff
@@ -0,0 +1,182 @@
...
+    )
+    assert result.type == ResultType.ACTION, result.result.message
+    assert "status" in result.result.data
+
+
+@pytest.mark.asyncio
+async def test_get_user_analytics(live_context):
+    result = await grammarly.execute_action(
+        "get_user_analytics",
+        {"date_from": "2025-12-01", "date_to": "2025-12-31"},
```
TheRealAgentK: Should fix: this hard-coded analytics date range will eventually age out of Grammarly's “past 365 days with 2-day lag” API window, making the live test fail over time. Please compute this dynamically, e.g. `date_to = today - 2 days` and `date_from = date_to - 30 days`, then format both as `YYYY-MM-DD`.

TheRealAgentK (commented):
Posted two inline review comments from the follow-up review.

After these are addressed, please also add the output of a new integration test run, @Shubhank-Jonnada

grammarly/grammarly.py
TheRealAgentK: A minimal version of the fix should be possible by passing a `yarl.URL` object through to `context.fetch()` (aiohttp accepts this even though the SDK annotation is `str`):

```python
from yarl import URL

async def upload_file(context: ExecutionContext, upload_url: str, file_content: str) -> bool:
    await context.fetch(
        URL(upload_url, encoded=True),
        method="PUT",
        data=file_content,
        content_type="text/plain",
    )
    return True
```

If we import `yarl` directly, please add it to `grammarly/requirements.txt` rather than relying on aiohttp/SDK transitive dependencies. If passing a non-`str` into `context.fetch()` is considered too loose, the fallback is to keep a dedicated aiohttp upload helper that does `session.put(URL(upload_url, encoded=True), ...)` for this one signed-URL upload path.

Shubhank-Jonnada: Fixed in https://github.com/Autohive-AI/autohive-integrations/pull/361/commits/b78f8de9f338431a6b61aaaa6e598436e91d245e . upload_file now wraps the pre-signed URL in URL(upload_url, encoded=True) from yarl to prevent aiohttp from re-encoding query parameters and invalidating the S3 signature. Added yarl>=1.9 to requirements.txt.

grammarly/tests/test_grammarly_integration.py
Shubhank-Jonnada: Fixed in a2d2e71. The analytics test now computes date_to as today minus 2 days and date_from as date_to minus 30 days, formatted as YYYY-MM-DD, so the range stays inside Grammarly's 365-day window indefinitely.

grammarly/tests/conftest.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+from unittest.mock import AsyncMock, MagicMock
+from autohive_integrations_sdk import ExecutionContext
+
+
+@pytest.fixture
+def mock_context():
+    ctx = MagicMock(spec=ExecutionContext)
+    ctx.fetch = AsyncMock()
+    ctx.auth = {
```
TheRealAgentK, line 10: Blocker: this mock auth shape is flat, so SDK 2.0.1 rejects it before the action handler runs (`VALIDATION_ERROR: flat auth is not supported`). Use the wrapped custom-auth envelope the platform passes, e.g. `{"auth_type": "Custom", "credentials": {"client_id": ..., "client_secret": ...}}`, and update the unit assertions to the current SDK 2.0.1 result shape instead of reading `result.result.data` if `result.result` is a dict in CI.

grammarly/requirements.txt
```diff
@@ -1 +1,2 @@
-autohive-integrations-sdk~=1.0.2
+autohive-integrations-sdk~=2.0.0
```
TheRealAgentK: Should fix: `autohive-integrations-sdk~=2.0.0` is now deprecated in the validator, and CI is already surfacing SDK 2.0.1 behavior. Please bump this to `~=2.0.1` or later and make the tests pass against that version.

grammarly/config.json
```diff
@@ -40,25 +40,20 @@
...
-            "type": "boolean",
-            "description": "Whether the operation was successful"
-          },
-          "error": {
-            "type": "string",
-            "description": "Error message if the action failed"
           }
         },
-        "required": ["result"]
+        "required": []
```
TheRealAgentK, line 56: Should fix: the config/code correctly removed the old in-band `result` / `error` fields, but `grammarly/README.md` still documents every action as returning `result` and `error` and shows old success/error examples. Please update the README to match the SDK v2 `ActionResult` / `ActionError` contract.

grammarly/config.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Grammarly",
   "display_name": "Grammarly",
-  "version": "1.0.0",
+  "version": "2.0.0",
```
TheRealAgentK, line 4: Should fix: this bumps the integration to 2.0.0, but the README version history still only lists 1.0.0. Please add a 2.0.0 entry covering the SDK v2 migration, `context.fetch()` migration, and `ActionError`/output-schema changes.

grammarly/tests/conftest.py
Shubhank-Jonnada: Fixed in c2995c9 (fix(grammarly): wrap mock auth in custom-auth envelope to match SDK 2.0.1 validation), the mock context now uses {"auth_type": "Custom", "credentials": {...}} matching what the platform actually passes, and the unit tests were re-run against real SDK 2.0.1 to confirm they pass.

grammarly/requirements.txt
Shubhank-Jonnada: Fixed in 2fd3643 (fix(grammarly): bump SDK pin to ~=2.0.1 and pass tests against it), bumped the pin and verified all 13 unit tests pass against the real 2.0.1 package.

grammarly/config.json
Shubhank-Jonnada: Fixed in 8278453 (docs(grammarly): document SDK v2 action result contract), removed the stale result/error lines from every action's Outputs and rewrote the Action Results section to describe the SDK v2 contract (data-only success, ActionError on failure).

grammarly/config.json
Shubhank-Jonnada: Fixed in 7b3228a (docs(grammarly): add 2.0.0 entry to README version history), added a 2.0.0 entry covering the SDK v2 migration.

Shubhank-Jonnada:
> Posted two inline review comments from the follow-up review.
> 
> After these are addressed, please also add the output of a new integration test run, @Shubhank-Jonnada 

image

@TheRealAgentK output from the new test run

TheRealAgentK (approved):
Re-reviewed after the latest SDK 2.0.1/custom-auth/test-contract fixes. The previously raised items are addressed: wrapped custom-auth shape, SDK 2.0.1 pin, SDK v2 README contract/version history, FetchResponse/ActionError handling, signed upload URL preservation, and integration-test contract. Unit validation is green and live-test evidence has been provided.

PR #362, opened 2026-06-10
feat: migrate supabase to SDK 2.0.0

Shubhank-Jonnada:
## Summary
- Bumps `autohive-integrations-sdk` to `~=2.0.0`
- Fixes auth to use flat `context.auth` (`host`, `service_role_secret`) instead of nested `credentials`
- Adds `.data` to all `context.fetch()` response accesses
- Converts all error returns from `ActionResult(data={"result": False})` to `ActionError(message=...)`
- Handles `delete_files` API-level error via `resp.data.get("error")`

## Tests
- Replaced `context.py` + `test_supabase.py` with `conftest.py`, `test_supabase_unit.py` (19 unit tests, all 15 actions covered), `test_supabase_integration.py`
- All 19 unit tests pass
- Integration tests skip without `SUPABASE_URL` / `SUPABASE_SERVICE_ROLE_KEY`


Shubhank-Jonnada:
All unit tests and live integration tests performed. 12/15 actions passed against the live API (3 skipped — `call_function` needs an RPC function, `get_user`/`delete_user` need auth users in the project).

supabase/supabase.py
```diff
@@ -193,42 +145,26 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
...
-            return ActionResult(
-                data={
-                    "records": result_records,
-                    "count": count,
-                    "result": True,
-                },
-                cost_usd=0.0,
-            )
-
+            return ActionResult(data={"records": result_records, "count": count}, cost_usd=0.0)
```
TheRealAgentK, line 165: Blocker: `delete_records` defaults to `return_records=False`, sets `count = None`, then returns `{"count": None}`. `config.json` declares `count` as an integer, so a normal delete without `return_records` will produce `VALIDATION_ERROR`. Please either allow `null` in the schema, return an integer count, or omit `count` when it is unknown.

supabase/tests/test_supabase_integration.py
```diff
@@ -0,0 +1,271 @@
...
+    )
+    result = await supabase.execute_action(
+        "delete_records",
+        {
+            "table": SUPABASE_TEST_TABLE,
+            "filters": {"id": f"eq.{record_id}"},
+        },
+        live_context,
+    )
+    assert result.type != ResultType.ACTION_ERROR, result.result.message
```
TheRealAgentK: Blocker: this misses the `delete_records` schema bug because `VALIDATION_ERROR` still passes this assertion. For successful live-path tests, please assert `result.type == ResultType.ACTION`.

supabase/tests/test_supabase_integration.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+# ---- Storage Actions (ordered: create → get → list → public_url → delete_files → delete_bucket) ----
+
+
+@skip_if_no_creds
+@pytest.mark.asyncio
+async def test_06_create_bucket(live_context):
```
TheRealAgentK: Blocker: these integration tests perform destructive operations without `@pytest.mark.destructive`, including insert, update, delete records, bucket create/delete, file delete, and user delete. Please mark destructive tests and document the safe command first:

```bash
pytest supabase/tests/test_supabase_integration.py -m "integration and not destructive"
```

Then document the destructive command separately with a warning.

.env.example
```diff
@@ -153,6 +153,10 @@
 # WHATSAPP_TEMPLATE_NAME=
 # WHATSAPP_TEMPLATE_LANG=
 # WHATSAPP_MEDIA_URL=
+# -- Supabase --
+# SUPABASE_URL=
+# SUPABASE_SERVICE_ROLE_KEY=
```
TheRealAgentK: Should fix: the integration tests also read `SUPABASE_TEST_TABLE`, `SUPABASE_TEST_FUNCTION`, and `SUPABASE_TEST_DELETE_USER_ID`. Please document those optional env vars here too.

supabase/tests/test_supabase_integration.py
```diff
@@ -0,0 +1,271 @@
...
+    records = [
+        {"id": str(uuid.uuid4()), "name": "integration-test-a"},
+        {"id": str(uuid.uuid4()), "name": "integration-test-b"},
+    ]
+    result = await supabase.execute_action(
+        "insert_records",
+        {"table": SUPABASE_TEST_TABLE, "records": records},
+        live_context,
+    )
+    assert result.type != ResultType.ACTION_ERROR, result.result.message
```
TheRealAgentK: Should fix: this pattern allows `VALIDATION_ERROR` to pass. Please use `assert result.type == ResultType.ACTION` for successful live-path tests.

supabase/tests/test_supabase_integration.py
```diff
@@ -0,0 +1,271 @@
...
+@skip_if_no_creds
+@pytest.mark.asyncio
+async def test_11_delete_files(live_context):
+    # Deleting nonexistent file — Supabase returns empty list or benign response
+    result = await supabase.execute_action(
+        "delete_files",
+        {"bucket_id": TEST_BUCKET, "paths": ["nonexistent/file.txt"]},
+        live_context,
+    )
+    assert result is not None
```
TheRealAgentK: Should fix: this can pass any failure wrapper. Please assert the expected `ResultType.ACTION`, or assert an explicit expected error if deleting a missing file is supposed to fail.

supabase/tests/test_supabase_integration.py
```diff
@@ -0,0 +1,271 @@
...
+                params=params,
+                json=json,
+                data=body,
+                headers=merged,
+            ) as resp:
+                try:
+                    data = await resp.json(content_type=None)
+                except Exception:
+                    data = await resp.text()
+                return FetchResponse(status=resp.status, headers=dict(resp.headers), data=data)
```
TheRealAgentK, line 75: Should fix: this returns `FetchResponse` for non-2xx statuses, unlike SDK `ExecutionContext.fetch()`, which raises. Several Supabase actions do not inspect `resp.status`, so invalid credentials or API errors can be misreported as successful empty results in live tests. Please raise `RateLimitError` for 429 and `HTTPError` for other non-2xx statuses.

supabase/tests/test_supabase_integration.py
TheRealAgentK: Suggestion: consider combining storage create, list, get, and delete into one destructive lifecycle test with cleanup in `finally`. That makes cleanup reliable even if later tests are skipped or someone runs one test by itself.

TheRealAgentK (changes requested):
Requesting changes for the Supabase SDK 2 migration blockers.

TheRealAgentK:

supabase/supabase.py
Shubhank-Jonnada: Fixed here - [https://github.com/Autohive-AI/autohive-integrations/commit/6899ec0](https://github.com/Autohive-AI/autohive-integrations/commit/6899ec0), returned `0` instead of `None` for `count` when `return_records=False`

supabase/tests/test_supabase_integration.py
Shubhank-Jonnada: Fixed here - [https://github.com/Autohive-AI/autohive-integrations/commit/4776205](https://github.com/Autohive-AI/autohive-integrations/commit/4776205), changed all assertions to `result.type == ResultType.ACTION`

supabase/tests/test_supabase_integration.py
Shubhank-Jonnada: Fixed here - [https://github.com/Autohive-AI/autohive-integrations/commit/22d42cb](https://github.com/Autohive-AI/autohive-integrations/commit/22d42cb), added `@pytest.mark.destructive` to all write operations and documented safe/destructive run commands

.env.example
Shubhank-Jonnada: Fixed here - [https://github.com/Autohive-AI/autohive-integrations/commit/b822ff8](https://github.com/Autohive-AI/autohive-integrations/commit/b822ff8), added `SUPABASE_TEST_TABLE`, `SUPABASE_TEST_FUNCTION`, and `SUPABASE_TEST_DELETE_USER_ID` to `.env.example`

supabase/tests/test_supabase_integration.py
Shubhank-Jonnada: Fixed here - [https://github.com/Autohive-AI/autohive-integrations/commit/4776205](https://github.com/Autohive-AI/autohive-integrations/commit/4776205), replaced all `!= ACTION_ERROR` with `== ResultType.ACTION`

supabase/tests/test_supabase_integration.py
Shubhank-Jonnada: Fixed here - [https://github.com/Autohive-AI/autohive-integrations/commit/22d42cb](https://github.com/Autohive-AI/autohive-integrations/commit/22d42cb), now asserts `result.type == ResultType.ACTION`

supabase/tests/test_supabase_integration.py
Shubhank-Jonnada: Fixed here - [https://github.com/Autohive-AI/autohive-integrations/commit/8acd867](https://github.com/Autohive-AI/autohive-integrations/commit/8acd867), `real_fetch` now raises `RateLimitError` on 429 and `HTTPError` on other non-2xx statuses

supabase/tests/test_supabase_integration.py
Shubhank-Jonnada: Fixed here - [https://github.com/Autohive-AI/autohive-integrations/commit/22d42cb](https://github.com/Autohive-AI/autohive-integrations/commit/22d42cb), combined storage tests into a single `test_storage_lifecycle` with `finally` cleanup

supabase/tests/test_supabase_integration.py
```diff
@@ -0,0 +1,279 @@
...
+    ]
+    result = await supabase.execute_action(
+        "insert_records",
+        {"table": SUPABASE_TEST_TABLE, "records": records},
+        live_context,
+    )
+    assert result.type == ResultType.ACTION, result.result.message
+    assert result.result.data["count"] >= 1
+
+    # Cleanup
```
TheRealAgentK: Nice cleanup work on the Supabase integration tests. One reliability tweak: this insert test does cleanup after the assertions, but not in a `finally`. If the insert succeeds and then an assertion or later cleanup call fails, test rows can be left behind. Could you wrap the cleanup loop in a `finally`, like the update/storage lifecycle tests do?

supabase/tests/test_supabase_unit.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+# ---- Database Actions ----
+
+
+@pytest.mark.asyncio
+async def test_select_records(mock_context):
+    mock_context.fetch = AsyncMock(return_value=MagicMock(data=[{"id": 1, "name": "Alice"}]))
+    result = await supabase.execute_action("select_records", {"table": "users"}, mock_context)
+    assert result.type != ResultType.ACTION_ERROR
```
TheRealAgentK: For happy-path unit tests, please prefer `assert result.type == ResultType.ACTION` instead of `!= ResultType.ACTION_ERROR`. The current pattern would still allow `VALIDATION_ERROR` to pass, which can hide schema/input drift in SDK v2 tests. This appears in several Supabase unit tests, so it’s worth updating the pattern throughout the file.

TheRealAgentK (commented):
Thanks for tightening up the Supabase test suite — the read-only/destructive split, env var documentation, non-2xx handling, and lifecycle-style storage test all look much better.

I left two smaller test-quality comments: make the insert cleanup robust with `finally`, and update happy-path unit assertions from `!= ACTION_ERROR` to `== ACTION` so SDK validation errors can’t accidentally pass. I’d also like to see a fresh live integration test run on the latest commit, since the earlier live-run note predates several of these fixes.

supabase/tests/test_supabase_integration.py
Shubhank-Jonnada: Fixed in commit `0073db8` - wrapped the cleanup loop in a `finally` block so DELETE runs even if the insert assertion fails, matching the pattern already used in `test_03_update_records` and `test_storage_lifecycle`.

supabase/tests/test_supabase_unit.py
Shubhank-Jonnada: Fixed in commit `af434ee` - replaced all 17 occurrences of `!= ResultType.ACTION_ERROR` with `== ResultType.ACTION` throughout `test_supabase_unit.py`. All 19 unit tests still passing.

supabase/tests/test_supabase_integration.py
```diff
@@ -0,0 +1,279 @@
...
+    not SUPABASE_TEST_TABLE,
+    reason="SUPABASE_TEST_TABLE required for database action tests",
+)
+
+
+@pytest.fixture
+def live_context(make_context):
+    service_role_key = SUPABASE_SERVICE_ROLE_KEY
+
+    async def real_fetch(url, *, method="GET", params=None, headers=None, json=None, body=None, **kwargs):
```
TheRealAgentK: Minor cleanup: `body` is accepted here and forwarded as `data=body`, but the Supabase source now sends request payloads with `json=` for the SDK 2 fetch path, so this kwarg appears to be dead in the live-test fixture. It is harmless as-is, but removing it would keep the fixture closer to the actual integration call sites and avoid implying that `body=` is part of the exercised contract.

TheRealAgentK (commented):
One minor cleanup note from re-review.

supabase/tests/test_supabase_integration.py
Shubhank-Jonnada: Fixed in `6739421` - removed body param from the signature and data=body from the request call.

Shubhank-Jonnada:
live integration test results:
Retested all the actions successfully, everything is working as expected

TheRealAgentK (approved):

PR #363, opened 2026-06-11
feat(bigquery): upgrade to SDK 2.0.0 with unit + integration tests

Tram-Nguyen87:
## Summary

Migrates the **Google BigQuery** integration from `autohive-integrations-sdk~=1.0.2` to `~=2.0.0`, following the Instagram migration as the reference and the `upgrading-sdk-v2` skill. All 14 actions updated, full unit + integration test coverage added.

## Source changes (`bigquery.py`)

- **FetchResponse**: every `context.fetch()` result is now read via `response.data` (the SDK 2.0.0 breaking change).
- **ActionError**: all error paths return `ActionError(message=...)` instead of `ActionResult(data={"result": False, "error": ...})`. The redundant `"result": True` success flag is dropped from every return.
- **Query error-body guard** (`run_query` / `get_query_results`): BigQuery can return **HTTP 200 with an `errors` body** when a query completes but fails at runtime — previously reported as an empty success. Now surfaced as `ActionError`, scoped to `jobComplete + errors + no rows + no schema` so non-fatal warnings on successful queries are never turned into false errors. (Mirrors the soft-error fix from the X migration; `insert_rows` and `get_job` already surface their in-body errors.)

## Config + dependencies

- `config.json`: version bumped to **2.0.0**; `error` / `result` properties removed from all output schemas; each `required` now points at a real output field.
- `requirements.txt`: pinned to `autohive-integrations-sdk~=2.0.0`.

## Tests

- **`tests/test_bigquery_unit.py`** — 57 FetchResponse-mocked unit tests: all 14 actions (success + `ActionError` + `VALIDATION_ERROR` paths), the `parse_rows`/`format_schema` helpers, and behavioural details (max-results capping, pagination tokens, camelCase time-partitioning conversion, partial insert failures, the query error-body guard incl. the warnings-still-succeed case).
- **`tests/test_bigquery_integration.py`** — `live_context` suite whose `real_fetch` returns `FetchResponse` and **raises `HTTPError` on non-2xx like the SDK**. Read-only tests (queries against public datasets, list/get datasets/tables/jobs/projects, dry-run, bad-query → `ACTION_ERROR`) plus a `@destructive` dataset→table→insert→delete lifecycle with `finally` cleanup. Gated behind the `integration` marker; skips without `BIGQUERY_ACCESS_TOKEN` / `BIGQUERY_PROJECT_ID`.
- Removed the old ad-hoc `tests/test_bigquery.py` and `tests/context.py`.

## Validation (local)

- `ruff check` ✅ and `ruff format --config ../autohive-integrations-tooling/ruff.toml` ✅
- **57 unit tests pass**; 13 integration tests collect & skip cleanly
- `validate_integration.py` ✅ (0 errors) and `check_code.py` ✅ (lint, format, bandit, pip-audit, config-code sync, fetch-pattern)

## Notes

- No `context.auth` lookup in this integration (platform OAuth2 — the SDK injects the bearer token), so the auth-default gotcha doesn't apply.
- Beta validation via a hand-built source zip is pending: such a bundle lacks the vendored SDK, so it isn't a substitute for a normal pipeline deploy — to be verified post-deploy.

Screenshot 2026-06-11 144634


Tram-Nguyen87:
Live E2E Test Results ✅
 Ran the integration suite against the real BigQuery REST API (project big-query-485807) with a live OAuth2 token.

bigquery/config.json
```diff
@@ -57,39 +60,46 @@
...
                         "description": "Whether the query result was served from cache"
-                    },
-                    "result": {
-                        "type": "boolean",
-                        "description": "Whether the operation was successful"
-                    },
-                    "error": {
-                        "type": ["string", "null"],
-                        "description": "Error message if the operation failed"
                     }
```
TheRealAgentK, line 108: 🚫 **Blocker:** `run_query` can now return `dry_run` from the dry-run path and `page_token` when BigQuery paginates results, but neither field is declared in this output schema. Success outputs should match `output_schema`; please add `dry_run` (`boolean`) and `page_token` (`["string", "null"]`) here.

bigquery/tests/test_bigquery_integration.py
```diff
@@ -0,0 +1,341 @@
+"""
+End-to-end integration tests for the BigQuery integration.
+
+These tests call the real BigQuery REST API and require:
+    BIGQUERY_ACCESS_TOKEN  — an OAuth2 access token with the BigQuery scope
+                             (https://www.googleapis.com/auth/bigquery)
+    BIGQUERY_PROJECT_ID    — a Google Cloud project ID you can query/write to
+
+Run all read-only tests:
+    pytest bigquery/tests/test_bigquery_integration.py -m integration
```
TheRealAgentK: ⚠️ **Should-fix:** this command also selects the destructive lifecycle test because the file has module-level `pytestmark = pytest.mark.integration` and the lifecycle class adds `@pytest.mark.destructive`. Please make the safe/read-only command `pytest bigquery/tests/test_bigquery_integration.py -m "integration and not destructive"`.

bigquery/tests/test_bigquery_integration.py
```diff
@@ -0,0 +1,341 @@
...
+
+# Make the integration importable as the ``bigquery`` package from the repo root.
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..")))
+
+from bigquery.bigquery import bigquery as bigquery_integration  # noqa: E402
+
+pytestmark = pytest.mark.integration
+
+ACCESS_TOKEN = os.environ.get("BIGQUERY_ACCESS_TOKEN", "")
+PROJECT_ID = os.environ.get("BIGQUERY_PROJECT_ID", "")
```
TheRealAgentK, line 35: ⚠️ **Should-fix:** this integration test reads `BIGQUERY_ACCESS_TOKEN` and `BIGQUERY_PROJECT_ID`, but root `.env.example` does not list either variable. Per the integration-test checklist, please add a BigQuery block there with blank template entries for both env vars.

bigquery/tests/test_bigquery_integration.py
```diff
@@ -0,0 +1,341 @@
...
+These tests call the real BigQuery REST API and require:
+    BIGQUERY_ACCESS_TOKEN  — an OAuth2 access token with the BigQuery scope
+                             (https://www.googleapis.com/auth/bigquery)
+    BIGQUERY_PROJECT_ID    — a Google Cloud project ID you can query/write to
+
+Run all read-only tests:
+    pytest bigquery/tests/test_bigquery_integration.py -m integration
+
+Run destructive tests (creates/deletes real datasets + tables in your project):
+    pytest bigquery/tests/test_bigquery_integration.py -m "integration and destructive"
```
TheRealAgentK, line 13: ⚠️ **Should-fix:** the README testing section still gives the generic `python -m pytest tests/` instruction and does not document the safe vs destructive integration-test commands. Since this PR adds a live destructive lifecycle test, please update `bigquery/README.md` to list the safe command first and the destructive command second with a mutation warning.

TheRealAgentK (changes requested):
Integration-specific review using the Autohive skills. I left one blocker for `run_query` output-schema drift and three should-fix comments for integration-test documentation/env-var setup.

bigquery/bigquery.py
```diff
@@ -556,19 +582,30 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
             for error in insert_errors:
                 formatted_errors.append({"index": error.get("index"), "errors": error.get("errors", [])})
 
+            # insertAll returns HTTP 200 even when rows are rejected. If nothing
+            # was inserted (e.g. an invalid row with the default skip_invalid_rows
+            # =False rejects the whole batch), surface it as a failure rather than
+            # an empty success. A partial insert (some rows landed) is returned as
+            # a normal result carrying inserted_count + insert_errors so callers
+            # see exactly what was written.
+            if insert_errors and inserted_count == 0:
```
TheRealAgentK: 🚫 **Blocker:** this condition is still too narrow for BigQuery's default insert semantics. BigQuery documents `skipInvalidRows=false` as causing the **entire request to fail if any invalid rows exist**, so a response with one `insertErrors` entry can still mean zero rows were inserted, not `len(rows) - failed_indices`.

The current fix only returns `ActionError` when every input row has an `insertErrors` entry. Please treat any `insertErrors` response as an `ActionError` when `skip_invalid_rows` is false, and only report partial inserted counts when `skip_invalid_rows=True`.

bigquery/tests/test_bigquery_integration.py
```diff
@@ -0,0 +1,341 @@
...
+                else:
+                    try:
+                        data = await resp.json(content_type=None)
+                    except Exception:
+                        data = await resp.text()
+                # Mirror the SDK: context.fetch() raises on non-2xx and only
+                # returns a FetchResponse for successful responses. Without this,
+                # error paths would never surface as ActionError in these tests.
+                if not resp.ok:
+                    raise HTTPError(resp.status, str(data), data)
```
TheRealAgentK, line 64: ⚠️ **Should-fix:** this fixture mirrors the SDK for generic non-2xx responses, but it still treats 429 rate limits as a plain `HTTPError`. Our integration-test pattern is to raise `RateLimitError` for 429 and `HTTPError` for other non-2xx statuses so live tests exercise the same SDK behavior as production.

Please import `RateLimitError` and branch 429 separately before this generic `HTTPError` path.

bigquery/tests/test_bigquery_integration.py
```diff
@@ -0,0 +1,341 @@
...
+
+# =============================================================================
+# PROJECTS
+# =============================================================================
+
+
+class TestListProjects:
+    async def test_returns_projects(self, live_context):
+        result = await bigquery_integration.execute_action("list_projects", {"max_results": 10}, live_context)
+        data = result.result.data
```
TheRealAgentK: ⚠️ **Should-fix:** live happy-path tests should assert `result.type == ResultType.ACTION` before dereferencing `result.result.data`. That makes `ACTION_ERROR` / `VALIDATION_ERROR` failures explicit and gives useful failure output via `result.result.message`.

This pattern appears throughout this integration file, not just this test. Please add the `ResultType.ACTION` assertion to each live success path before reading `result.result.data`.

bigquery/tests/test_bigquery_integration.py
```diff
@@ -0,0 +1,341 @@
...
+
+@pytest.mark.destructive
+class TestDatasetTableLifecycle:
+    """Create a dataset + table, stream a row, then tear everything down.
+
+    Cleans up after itself even if an assertion fails midway.
+    """
+
+    async def test_full_lifecycle(self, live_context, project_id):
+        dataset_id = "autohive_it_dataset"
```
TheRealAgentK: ⚠️ **Should-fix:** this hardcoded dataset ID makes destructive live runs fragile. Concurrent runs will collide, and a stale leftover from a failed run can make the next run fail before it proves anything about the integration.

Please generate a unique dataset/table suffix for each run, e.g. with `uuid.uuid4().hex[:8]` or similar, while keeping the `autohive_it_` prefix for easy identification.

bigquery/tests/test_bigquery_integration.py
```diff
@@ -0,0 +1,341 @@
...
+        table_id = "autohive_it_table"
+
+        # create dataset
+        create_ds = await bigquery_integration.execute_action(
```
TheRealAgentK: ⚠️ **Should-fix:** the lifecycle docstring says cleanup runs even if an assertion fails, but `create_dataset` happens before the `try/finally`. If the dataset is created and this assertion (or any code before the `try`) fails, cleanup will not run.

Please move dataset creation inside the `try` and make the `finally` cleanup best-effort/guarded, so any created dataset is deleted even when an early assertion fails.

TheRealAgentK (changes requested):
Integration-specific re-review. I left one blocker for BigQuery `insert_rows` failure semantics and several should-fix comments on live integration-test quality/safety.

Tram-Nguyen87:
Addressed in 1c639f3:
  -  insert_rows semantics: with skip_invalid_rows=False (default), any insertErrors now returns ActionError (BigQuery writes zero rows when any row is invalid). Partial inserted_count is only reported when
  skip_invalid_rows=True; all-rejected there is still an error. Unit tests cover both modes.
  - Live tests: 429 → RateLimitError (others → HTTPError); assert ResultType.ACTION before reading .data; uuid suffix on the lifecycle dataset/table; create_dataset moved inside try with best-effort finally cleanup.
  All checks + 59 unit tests pass locally.

TheRealAgentK:
Please resolve the conflicts and run fresh integration test and post the results.

Tram-Nguyen87:
**Re-review ready — conflict resolved + all review comments addressed.**

**Merge conflict:** resolved the `.env.example` conflict — added the Google BigQuery block in its alphabetical slot (Google cluster), kept master's Zoom entry, and removed duplicate Canva/X/Microsoft 365 lines that had snuck in. PR is now mergeable.

**Round-1 comments:**
- ✅ `run_query` output schema now declares `dry_run` (boolean) and `page_token` (`["string","null"]`).
- ✅ Read-only command corrected to `-m "integration and not destructive"` (docstring + README).
- ✅ Added BigQuery block to root `.env.example` (`BIGQUERY_ACCESS_TOKEN`, `BIGQUERY_PROJECT_ID`).
- ✅ README testing section lists the safe command first, destructive second with a mutation warning.

**Round-2 comments:**
- ✅ `insert_rows`: any `insertErrors` response now returns `ActionError` when `skip_invalid_rows=False` (entire-request-fails semantics); partial inserted counts only reported when `skip_invalid_rows=True`.
- ✅ Live fetch fixture raises `RateLimitError` for 429, `HTTPError` for other non-2xx.
- ✅ All live success paths assert `result.type == ResultType.ACTION` (via `_action_data` helper) before reading `.data`.
- ✅ Destructive lifecycle uses a `uuid` suffix per run; `create_dataset` is inside the `try`, cleanup in a guarded best-effort `finally`.

**Fresh test runs (today):**
- Unit: `59 passed`
- Integration (safe): `12 passed` against live BigQuery
- Integration (destructive lifecycle): `1 passed` against live BigQuery — created and deleted a real dataset/table; verified no leftovers.

TheRealAgentK (approved):

PR #368, opened 2026-06-12
feat(canva): upgrade to SDK 2.0.0 with unit + integration tests

Tram-Nguyen87:
## Summary

Migrates the **Canva** integration from `autohive-integrations-sdk` **1.0.2 → 2.0.0**, following the `upgrading-sdk-v2` skill and using the already-migrated Instagram integration as the reference pattern.

The only breaking SDK change is that `context.fetch()` now returns a `FetchResponse` (`.status` / `.headers` / `.data`) instead of the parsed body directly. This PR also adopts `ActionError` (introduced in SDK 1.1.0, never picked up on 1.0.x) for all error paths.

### Source — `canva.py`
- Read the response body via `FetchResponse.data` at every fetch call site (connected-account handler + all 18 body-reading actions). The 3 write-only calls (`delete_*`, `update_*`, `move`) don't read a body, so they're unchanged.
- Convert **every** error path from `ActionResult(data={"result": False, "error": str(e)})` to `return ActionError(message=str(e))`; add `ActionError` to the SDK import.
- Drop the vestigial `"result": True` success flag — success/failure is now expressed by the result type (`ACTION` vs `ACTION_ERROR`), not an in-payload boolean.
- Read optional inputs via `inputs.get("x")` (matching the Instagram idiom) instead of `inputs["x"]` inside `if "x" in inputs:` guards. Required params keep `inputs["x"]`. Behavior is preserved exactly; this clears the config-code-sync checker's false-positive KeyError warnings.

### `config.json`
- Remove the `"result"` and `"error"` properties from all **21** action output schemas, and drop `"result"` from each `required` array (`["designs"]` / `["items"]` preserved).
- Bump integration `version` → `2.0.0`.

### `requirements.txt`
- Pin `autohive-integrations-sdk~=2.0.0`.

### Tests
- **`tests/test_canva_unit.py`** — 71 mocked unit tests: `FetchResponse`-wrapped fetch mocks, `ResultType.ACTION_ERROR` / `ResultType.VALIDATION_ERROR` assertions, branch + payload-construction coverage (upload/import file handling, export format options, pagination/continuation), connected-account name splitting, and a **parametrized error-path sweep across all 21 actions** verifying each `except` block now returns `ActionError`.
- **`tests/test_canva_integration.py`** — `live_context` fixture whose `real_fetch` returns `FetchResponse` (Bearer auth from `CANVA_ACCESS_TOKEN`). Read-only tests plus design/folder/asset lifecycle tests gated behind `@pytest.mark.destructive`.
- Replaced the old standalone `tests/test_canva.py` + `tests/context.py` (a manual script with a hard-coded token) with the pytest unit/integration split + a `conftest.py` path shim.
- Removed `canva/__init__.py` — the empty file made `canva/` a package that shadowed the `canva.py` module and broke `from canva import ...` in tests. Matches Instagram's single-file layout; the Lambda entry point (`entry_point: "canva.py"`) is unaffected.

## Validation
- `ruff check` + `ruff format --config ../autohive-integrations-tooling/ruff.toml` — clean
- **71 unit tests pass**; 8 integration tests collect and skip cleanly without a token
- `check_code.py` — **PASSED** with **no warnings** (syntax, imports, JSON, lint, format, bandit, pip-audit, config-code sync, fetch patterns)
- `validate_integration.py` — **0 errors**, 2 expected warnings

The 2 remaining `validate_integration.py` warnings are both expected and benign:
1. *Missing `__init__.py`* — the deliberate package-shadowing fix described above.
2. *`profile:read` possibly unused* — false positive; the scope is used by the connected-account handler (`/v1/users/me/profile`). Pre-existing.

## Behavior note for reviewers
Successful action outputs no longer include the `"result": true` boolean (and failures no longer carry `"result": false` / `"error"`). Errors now surface as `ActionError`, which bypasses output-schema validation — consistent with the SDK 2.0.0 contract and the Instagram migration. Downstream agents that branched on the `result` field should rely on the action erroring instead.

image

image


Tram-Nguyen87:
@codex review please

Tram-Nguyen87:
Canva e2e tests — all passed ✅
  - Read-only: 5/5 passed
  - Destructive: 3/3 passed (folder + asset lifecycles self-cleaned)
  - Total: 8/8 against the live Canva API, covering all 18 actions

TheRealAgentK:
Nice work on this SDK 2.0.0 migration, Tram, the code/test migration is in good shape and the local validation/unit tests are green on my pass.

I found two small should-fixes before merge:

1. The integration-test docstring labels this as the read-only command:
   ```bash
   pytest canva/tests/test_canva_integration.py -m integration
   ```
   Because this file also has `@pytest.mark.destructive` tests, that command will run the destructive design/folder/asset lifecycle tests too. Please change the safe/read-only command to:
   ```bash
   pytest canva/tests/test_canva_integration.py -m "integration and not destructive"
   ```

2. `canva/tests/test_canva_integration.py` reads `CANVA_ACCESS_TOKEN`, but root `.env.example` does not list it yet. Please add a Canva block there, e.g.:
   ```bash
   # -- Canva --
   # CANVA_ACCESS_TOKEN=
   ```

TheRealAgentK:
Sorry, one more should-fix I missed from my notes:

The destructive folder and asset lifecycle tests clean up only at the very end. If an assertion or API call fails after creating the folder/asset, the test can leave real Canva resources behind. Please wrap cleanup in `try/finally` once the created resource ID is known, where a delete action exists:

- `TestFolderLifecycle.test_full_lifecycle`: delete the folder in `finally` after `folder_id` is set.
- `TestAssetLifecycle.test_full_lifecycle`: delete the asset in `finally` after `asset_id` is set.

The design lifecycle does not appear to have a delete action available here, so this mainly applies to folder/asset cleanup.

Tram-Nguyen87:
Thanks for the review, Kai — all three should-fixes are in (commit `fdd0f4a`):

1. **Read-only run command** — docstring now uses `pytest canva/tests/test_canva_integration.py -m "integration and not destructive"` so the safe command no longer pulls in the destructive lifecycle tests.
2. **`.env.example`** — added a `# -- Canva --` block with `# CANVA_ACCESS_TOKEN=` at the repo root.
3. **Destructive cleanup** — `TestFolderLifecycle` and `TestAssetLifecycle` now delete the created folder/asset in a `try/finally` once the ID is known, so resources are cleaned up even if a mid-lifecycle assertion fails. Left the design lifecycle as-is since there's no delete action for designs.

Also merged latest `master` to resolve the `.env.example` conflict (the Microsoft 365 migration appended to the same end-of-file spot) — both blocks are kept. Ready for re-review 🙏

TheRealAgentK (dismissed):
## Integration PR re-review

Applied sub-skills: building-integration, writing-unit-tests, writing-integration-tests, upgrading-sdk-v2

### 🚫 Blockers
None.

### ⚠️ Should-fix
None. The prior review items are addressed:
- Read-only integration-test command now uses `-m "integration and not destructive"`.
- Root `.env.example` documents `CANVA_ACCESS_TOKEN`.
- Folder and asset destructive integration tests now clean up with `try/finally` once the resource ID is known.

### 💡 Suggestions
None.

### Validation
- CI is green for Validate PR title, Validate Integration, and Analyze (python).
- Local tooling run:
  - `validate_integration.py canva` — passed with the two expected warnings (`__init__.py` intentionally absent; `profile:read` used by connected-account profile lookup).
  - `check_code.py canva` — passed.
  - `run_tests.py canva` — 71/71 passed, 94% coverage.
  - `check_readme.py origin/master canva` — passed.
  - `check_version_bump.py origin/master canva` — passed (`1.0.0 → 2.0.0`).
- Local tests run:
  - `pytest canva/tests/test_canva_unit.py -q` — 71 passed.
  - `pytest canva/tests/test_canva_integration.py -m "integration and not destructive" -q` — 5 skipped without `CANVA_ACCESS_TOKEN`, 3 destructive tests deselected.
- SDK v2 checks inspected: all action decorators match config actions, fetch body reads use `.data`, errors return `ActionError`, SDK pin is `autohive-integrations-sdk~=2.0.0`, and output schemas no longer expose error-only `result`/`error` fields.
- Security/public-safety: no Canva-scoped secret patterns found in `.env.example` or `canva/`; `check_code.py` bandit/pip-audit passed.

Looks good to merge.

Tram-Nguyen87 (commented):
## Integration PR review — SDK 1.0.2 → 2.0.0

Applied sub-skills: `upgrading-sdk-v2`, `building-integration`, `writing-unit-tests`, `writing-integration-tests`.

Reviewed the full diff against `origin/master` (canva.py, config.json, requirements.txt, tests, `.env.example`), the SDK 2.0.0 `fetch` contract, and the pre-migration source to confirm behaviour preservation.

### 🚫 Blockers
None.

### ⚠️ Should-fix
None.

### 💡 Suggestions (all pre-existing, out of scope for this SDK bump — noting for a future cleanup, do **not** block on these)
1. **`import_design` builds its `Import-Metadata` header with `str(metadata).replace("'", '"')`** (`canva.py:420`). This is a dict-repr → pseudo-JSON hack that survives only because the values are base64/MIME with no quotes or non-ASCII. `upload_asset` (`canva.py:97`) hand-rolls the same header via an f-string. Both would be safer as `json.dumps(...)`. Pre-existing on `master`; the migration preserved it verbatim, so it's correctly out of scope here.
2. **`create_folder` emits `{"folder": None}` when the response lacks a `folder` key** (`canva.py:523`), whereas every other read action guards with `if response.data.get(...)`. Harmless in practice (create always returns a folder) and behaviour-identical to `master`, but the inconsistency is worth tidying next time the file is touched.
3. **`config.json` describes `image_quality` as applying to "PNG/MP4/GIF"** (`config.json:425`), but the code only consumes it in the MP4/GIF branch — the PNG branch ignores it (uses `width`/`height`/`lossless`). I verified this matches `master` exactly (the PNG branch never read `image_quality`), so it's a stale doc string, not a regression. Tighten the description when convenient.

### Verification

**SDK 2.0 contract — `fetch` error semantics (the part that most often bites a migration):**
- `context.fetch` raises `HTTPError` on any non-2xx (`integration.py:520-522`) and `RateLimitError` on 429 (`:498-505`). Both subclass `Exception`.
- Canva's APIs signal failure through HTTP status, so every `except Exception: return ActionError(message=str(e))` block genuinely catches upstream failures — there is **no "HTTP 200 but failed in the body" path** to leak through as a false success. The soft-error audit is clean for this integration.
- `.data` is read at every body-reading call site; the 3 write-only calls (`update_*`, `delete_*`, `move`) correctly don't touch the body.

**Behaviour preservation (diffed each action against `master`):**
- Error paths converted from `ActionResult(data={"result": False, "error": ...})` → `ActionError(message=...)` — the intended v2 contract change; success/failure now lives in the result *type*.
- Vestigial `"result": True` dropped from success payloads, and the matching `result`/`error` properties removed from all 21 output schemas (+ `required` arrays trimmed to `["designs"]`/`["items"]` where applicable).
- `inputs[...]`-inside-`if "x" in inputs` rewritten to `inputs.get("x")` — payload construction is byte-identical (I checked `export_design`'s full format-branch logic, the `upload_asset`/`import_design` file-vs-files fallback, and the pagination/continuation handling).
- 21 `@canva.action` decorators ↔ 21 config actions: exact 1:1 match.

**Tooling (local):**
- `pytest canva/tests/test_canva_unit.py -q` → **71 passed** in 0.30s.
- CI report on the merge commit: Structure ⚠️ (2 expected warnings — intentional absent `__init__.py`; `profile:read` *is* used by the connected-account profile lookup), Code ✅, Tests ✅ **71/71, 94%**, README ✅, Version ✅ `1.0.0 → 2.0.0`.
- `requirements.txt` pinned `autohive-integrations-sdk~=2.0.0`; `.env.example` documents `CANVA_ACCESS_TOKEN`.

**Tests:**
- Unit suite mocks `fetch` to return `FetchResponse` and asserts `ResultType.ACTION` / `ACTION_ERROR` / `VALIDATION_ERROR`. Strong coverage: connected-account name splitting (single / multi-word / empty), upload & import file-handling branches, every export format branch, pagination, and a **parametrized error-path sweep across all 21 actions** proving each `except` now returns `ActionError` — exactly the guard against "rewrote most blocks but missed one".
- Integration suite: `live_context` whose `real_fetch` returns a real `FetchResponse` (Bearer from `CANVA_ACCESS_TOKEN`), read-only tests plus design/folder/asset lifecycles gated behind `@pytest.mark.destructive`, with `try/finally` cleanup once the resource ID is known. Note: `real_fetch` raises a bare `Exception` (not `HTTPError`) on ≥400 — fine since the handlers catch `Exception`, just slightly looser than the real SDK type.
- Replacing the hard-coded-token `test_canva.py`/`context.py` script with the unit/integration split + `conftest.py` shim, and dropping the package-shadowing `__init__.py`, are both correct and match the Instagram reference layout.

### Verdict
Faithful, behaviour-preserving SDK 2.0.0 migration with full unit + integration coverage and a clean soft-error audit. ✅ Looks good to merge. The three suggestions are pre-existing and explicitly out of scope.

TheRealAgentK:
I approved and merged this - please fix up the README file (version of the integration, documentation still pointing to SDK 1.x in a new follow-up PR today, @Tram-Nguyen87

TheRealAgentK (approved):

PR #369, opened 2026-06-12
feat(calendly): upgrade to SDK 2.0.0 with unit + integration tests

Tram-Nguyen87:
## Summary

Upgrades the **Calendly** integration from `autohive-integrations-sdk~=1.0.2` to `~=2.0.0`, following the `upgrading-sdk-v2` skill, with the migrated Instagram integration as the reference for test structure.

All 20 actions migrated; action behaviour, inputs, and outputs are unchanged (other than dropping the internal `result` flag).

## Changes

**Source (`calendly.py`)**
- Access response bodies via `FetchResponse.data` at all 20 `context.fetch()` sites (e.g. `response.data.get("resource", response.data)`).
- Convert every error path from `ActionResult(data={"result": False, "error": ...})` to `return ActionError(message=str(e))` — returned, not raised.
- Drop the redundant `"result": True` success flag (success/error now signalled by `ResultType`).

**Source cleanup (`calendly.py`)**
- Replace the dynamic `for key in [...]: inputs[key]` param loops in the five list actions with explicit per-parameter `inputs.get(...)` blocks, so the config-code sync checker can see each input is accessed (clears all "parameter defined in input_schema but never accessed" warnings). Behaviour is identical — `is not None` guards are preserved (so a falsy-but-present value like `active=False` is still forwarded), and `list_webhooks` accesses its required `organization` directly, matching the existing `list_routing_forms` pattern.

**Config & deps**
- `config.json`: version `1.0.0 → 2.0.0`; removed the error-only `"result"` property from all 20 output schemas.
- `requirements.txt`: pinned `autohive-integrations-sdk~=2.0.0`.
- `README.md`: added a 2.0.0 Version History entry.

**Tests** (replaces the old script-style `test_calendly.py`)
- `test_calendly_unit.py` — **65** mocked tests; every action covered for success, `ActionError` (fetch raises → `ResultType.ACTION_ERROR`), and `ResultType.VALIDATION_ERROR` (missing required input). Plus param-forwarding, URL shape, the `routing_form → form` remap, body construction, and `resource` fallback.
- `test_calendly_integration.py` — **13** live tests behind the `integration`/`destructive` markers. `real_fetch` **raises `HTTPError`/`RateLimitError` on non-2xx**, faithfully mirroring the SDK contract so an error response surfaces as `ActionError` instead of masquerading as success data.
- `tests/context.py` → `tests/conftest.py` (path bootstrap), matching the Instagram layout.

## Contract verification

Confirmed against the installed SDK source (`integration.py`) that `context.fetch()` still **raises** `HTTPError`/`RateLimitError` on non-2xx and only returns a `FetchResponse` for 2xx — so `.data` is only ever read on success and the `try/except → ActionError` pattern is correct. The two write-only fetches (`cancel`, `delete`) don't read `.data`, so a 204 (`data=None`) can't fault.

## Validation (all green locally)

- `ruff check` + `ruff format` (tooling config, line-length 120)
- `pytest test_calendly_unit.py` — 65 passed
- `validate_integration.py` — 0 errors / 0 warnings
- `check_code.py` — CODE CHECK PASSED (syntax, imports, JSON, lint, format, bandit, pip-audit, config-sync **with no warnings**, fetch-pattern)
- `check_readme.py` — README CHECK PASSED

## Not done in this PR

Integration tests are written and collect cleanly but have **not been run against the live Calendly API** (requires `CALENDLY_ACCESS_TOKEN`; webhook/routing-form tests additionally need a paid plan). They skip gracefully without a token.


TheRealAgentK:
Good SDK 2.0.0 migration overall — local validation and unit tests are green on my pass.

I found the same small should-fix pattern as the Canva PR:

1. The integration-test docstring labels this as the read-only command:
   ```bash
   pytest calendly/tests/test_calendly_integration.py -m integration
   ```
   Because this file also has a `@pytest.mark.destructive` webhook lifecycle test, that command will run the destructive create/delete test too. Please change the safe/read-only command to:
   ```bash
   pytest calendly/tests/test_calendly_integration.py -m "integration and not destructive"
   ```

2. `calendly/tests/test_calendly_integration.py` reads `CALENDLY_ACCESS_TOKEN`, but root `.env.example` does not list it yet. Please add a Calendly block there, e.g.:
   ```bash
   # -- Calendly --
   # CALENDLY_ACCESS_TOKEN=
   ```

3. The destructive webhook lifecycle test deletes only at the very end. If an assertion or API call fails after creating the webhook, the test can leave a real webhook subscription behind. Please wrap deletion in `try/finally` once the webhook UUID is known.

Tram-Nguyen87:
## e2e testing — done ✅
 Ran against the **live Calendly API** with a real token, plus unit tests.
  - **Unit:** 65 passed
  - **Integration (read-only):** 11 passed, 1 skipped

  **Skipped (1):** `test_get_event_and_invitees` — by design, test account has no scheduled events (auto-skips). Destructive webhook test not run (paid plan, opt-in).
  **Bug found + fixed (`8657d0c`):** `list_event_types` `active` filter was broken — schema is boolean but aiohttp can't send a bool as a query param and Calendly  wants `"true"/"false"`. Fixed with `str(inputs["active"]).lower()` + updated the unit test.

TheRealAgentK (approved):
## Integration PR re-review

Applied sub-skills: building-integration, writing-unit-tests, writing-integration-tests, upgrading-sdk-v2.

### 🚫 Blockers
None.

### ⚠️ Should-fix
None.

### 💡 Suggestions
None.

### Validation
- CI checked: latest PR head is green for Validate Integration, Validate PR title, and Analyze (python).
- Local tooling run:
  - `.venv/bin/python ../autohive-integrations-tooling/scripts/validate_integration.py calendly` ✅
  - `.venv/bin/python ../autohive-integrations-tooling/scripts/check_code.py calendly` ✅
  - `.venv/bin/python ../autohive-integrations-tooling/scripts/run_tests.py calendly` ✅ — 65/65, 93% coverage
  - `.venv/bin/python ../autohive-integrations-tooling/scripts/check_readme.py origin/master calendly` ✅
  - `.venv/bin/python ../autohive-integrations-tooling/scripts/check_version_bump.py origin/master calendly` ✅ — 1.0.0 → 2.0.0 major bump
  - `.venv/bin/python -m pytest calendly/tests/test_calendly_unit.py -q` ✅ — 65 passed
  - `.venv/bin/python -m pytest calendly/tests/test_calendly_integration.py -m "integration and not destructive" -q` ✅ — skips cleanly without `CALENDLY_ACCESS_TOKEN`
- Re-checked prior review items: safe read-only integration-test command is documented, `CALENDLY_ACCESS_TOKEN` is listed in root `.env.example`, and webhook lifecycle cleanup now uses `try/finally` after UUID capture.
- Security/public-safety spot check: no real Calendly token or committed credential found; test URLs and emails are placeholders.

PR #370, opened 2026-06-15
feat(circle): migrate to autohive-integrations-sdk 2.0.0

Tram-Nguyen87:
## Summary

Migrates the **Circle** integration from `autohive-integrations-sdk` 1.x to **2.0.0**, adopting the two breaking/behavioural changes the SDK introduced:

- **`FetchResponse`** — `context.fetch()` now returns a `FetchResponse` object, so every call site reads the body via `response.data` (and `response.status` where we need to detect non-2xx).
- **`ActionError`** — all error paths now `return ActionError(message=...)` instead of `ActionResult(data={"error": ...})`, so failures skip output-schema validation and surface cleanly. The `"error"`/`"result"` signalling fields were removed from the affected output schemas in `config.json`.

`requirements.txt` pins `~=2.0.0` and `config.json` is bumped to `2.0.0` (major bump for the breaking SDK dependency).

## What changed

- `circle.py` — `.data` added to all 23 fetch call sites; all 24 error returns converted to `ActionError`; shared `handle_api_response()` helper updated to use `.data` / `.status`.
- Auth path is crash-safe: `build_auth_headers()` is called **inside** each action's `try`, so a missing/invalid token returns a clean `ActionError` rather than crashing the Lambda.
- `config.json` — version bump + removal of error-signalling output fields.
- Tests restructured to the standard layout: `test_circle_unit.py` + `test_circle_integration.py` (replacing the old single `test_circle.py`).

## Testing

- **Unit:** 48 tests, **all passing** — covers **all 23 actions** (FetchResponse-wrapped mocks, `ResultType.ACTION_ERROR` / `VALIDATION_ERROR` assertions).
- **Integration (e2e):** read-path + destructive-lifecycle coverage for **16/23 actions** via the `live_context` / `real_fetch → FetchResponse` pattern (matches the Instagram migration). Includes a `@pytest.mark.destructive` post create→update lifecycle. Run locally with `CIRCLE_API_TOKEN`:
  ```
  pytest circle/tests/test_circle_integration.py -m "integration and not destructive"
  ```
  The 7 actions without e2e coverage are all destructive membership/tag/group mutations with no inverse cleanup action — deliberately left out to avoid leaving artifacts in the community.
- **Beta:** packaged with `hiveup package` and **deployed + tested successfully on beta**.

## Note

Includes a small `.gitignore` housekeeping commit (ignore root-level `*.zip` build artifacts and `.pr_body_*.md` scratch files) so `hiveup package` output can't be committed by accident.

Screenshot 2026-06-15 153638

Screenshot 2026-06-15 153648


Tram-Nguyen87:
closed due to migration has been taken by other.

PR #372, opened 2026-06-16
feat(zoom): upgrade to SDK 2.0.0 with unit + integration tests

Tram-Nguyen87:
## Summary

Migrates the **Zoom** integration from `autohive-integrations-sdk` 1.0.2 to 2.0.0, adopting the `FetchResponse` and `ActionError` contracts, and replaces the old script-style tests with full mocked unit + live integration suites.

## Source (`zoom.py`)

- **FetchResponse:** Unwrap `.data` *inside* the central `_make_request` helper (the only place `context.fetch()` is called). This preserves the helper's existing contract — it still returns the dict body — so all 25 action call sites are unchanged and there's no per-site `.data` rewriting to get subtly wrong.
- **DELETE fix:** Changed the empty-body branch from `response if response else {...}` to `response.data if response.data else {"success": True}`. A `FetchResponse` is always truthy, so without this the empty-body branch would never fire — a real latent bug.
- **ActionError:** Every error path now returns `ActionError(message=str(e))` instead of `ActionResult` with `error`/`result` fields (25 actions).
- Dropped the now-redundant internal `"result": True` success flag.

## Config & deps

- Bumped `config.json` version → `2.0.0`.
- Removed the error-only `"result"` / `"error"` properties from all 25 output schemas and trimmed `"result"` from their `required` arrays (input schemas untouched).
- Pinned `autohive-integrations-sdk~=2.0.0`.
- Populated `__init__.py` to re-export the integration (it was an empty 0-byte file that shadowed `zoom.py` during test import).

## Tests

- **`test_zoom_unit.py` — 81 tests, all passing.** Every one of the 25 actions has success / `ActionError` / validation-error coverage, plus the connected-account handler, helper functions (`encode_meeting_id`, `get_headers`), `_make_request` behaviour (body unwrap, DELETE-empty marker, unknown-method `ValueError`), and input-boundary checks.
- **`test_zoom_integration.py`** — live read-only tests + a destructive create→get→update→delete meeting lifecycle with guaranteed cleanup. `real_fetch` raises `HTTPError`/`RateLimitError` on non-2xx to mirror the SDK, so error responses surface as `ActionError` rather than masquerading as success data. Gated on `ZOOM_ACCESS_TOKEN`; never runs in CI.

## Validation

All local CI checks pass: `ruff check`, `ruff format` (tooling config), `validate_integration.py` (only a pre-existing unused-scope warning for `zoomapp:inmeeting`), and `check_code.py` (bandit, pip-audit, config-code sync, fetch-pattern linter all ✅).

image


zoom/tests/test_zoom_integration.py
```diff
@@ -0,0 +1,171 @@
...
+        assert isinstance(result.result.data["contacts"], list)
+
+
+# =============================================================================
+# DESTRUCTIVE — create, update, then delete a real meeting
+# Only run with: pytest -m "integration and destructive"
+# =============================================================================
+
+
+@pytest.mark.destructive
```
TheRealAgentK, line 144: ⚠️ Should-fix: These integration tests cover the core user/meeting paths and a meeting lifecycle, but this integration exposes 25 actions. Per our integration-test skill for large integrations, we should either add live coverage for the other read-only/key write domains (calendar actions, settings, templates, invite links, participants) or call out in the PR why this PR intentionally limits live coverage to the meeting workflow.

zoom/config.json
TheRealAgentK, line 1: ⚠️ Should-fix: Local/CI structure validation still warns that `zoomapp:inmeeting` is potentially unused. Please verify this scope is required by one of the implemented actions; if it is not needed, remove it to keep the OAuth scope set minimal.

TheRealAgentK (commented):
Reviewed with the repo integration skills. No blockers found; left two should-fix comments inline/file-level: integration-test coverage breadth and the remaining potentially-unused Zoom scope warning.

Tram-Nguyen87:
**Live e2e regression ✅**

Ran the full integration suite against the real Zoom API v2 (read-only + destructive):

```
pytest zoom/tests/test_zoom_integration.py -m "integration" -v
→ 7 passed in 4.38s
```

Covers connected account, get user, permissions, list meetings/contacts, not-found → clean `ACTION_ERROR`, and the destructive create→update→delete lifecycle (cleanup guaranteed in `finally`, no leftover data).

TheRealAgentK:
Did you want to address the two comments, @Tram-Nguyen87 ?

Tram-Nguyen87:
Thanks @TheRealAgentK — addressed both comments:

**1. `zoomapp:inmeeting` scope** — Removed (commit be69054). It belongs to Zoom's in-client Zoom Apps SDK and isn't used by any of the 25 REST actions, so it was pure over-provisioning. Scope set is now minimal (24) and the structure validator warning is cleared.

**2. Integration-test breadth** — All 25 actions are covered by unit tests (82 tests, 93%). I kept the live tests focused on the user/meeting workflow on purpose: the other actions need things a normal test account doesn't have — Zoom Calendar provisioned, a meeting that's already *ended* for participant data, a registration-enabled meeting — so live tests for them would just skip rather than actually verify anything. Happy to add skip-when-unavailable live tests for the read-only paths in a follow-up if you'd prefer.

TheRealAgentK (approved):

PR #374, opened 2026-06-17
feat(zoho): upgrade to SDK 2.0.0 with unit + integration tests

Tram-Nguyen87:
## Summary

Upgrades the Zoho CRM integration to `autohive-integrations-sdk~=2.0.0`, with a full unit + live integration test suite.

## Changes

- **SDK 2.0.0 migration** — access `FetchResponse.data` at every `context.fetch()` call site across all 54 actions (single- and multi-fetch paths).
- **Error handling** — convert every error path to `ActionError(message=...)`, including the Zoho soft errors (non-`SUCCESS` response code, empty `data` envelope). Dropped the vestigial `"result": True` from success returns.
- **Dependencies & config** — `requirements.txt` → `autohive-integrations-sdk~=2.0.0`; `config.json` bumped to `2.0.0` and the now-redundant `result`/`error` output-schema props removed from all actions.
- **Tests** — added a pytest unit suite (209 tests covering all 54 actions) and a live integration suite (read-only + destructive lifecycles). Removed the old manual `test_zoho.py` script.

## Test plan

- [ ] `pytest zoho/tests/test_zoho_unit.py` — 209 unit tests pass
- [ ] Live integration suite against a real Zoho CRM sandbox (read-only + destructive lifecycles)
- [ ] Build/validation pipeline passes
 
image

image


Tram-Nguyen87:
### Config-code sync warnings

The local build pipeline passes all gates. The `check_config_sync` step emits two classes of **warnings** (it returns exit 0 — non-blocking):

1. **`will raise KeyError` (21) — fixed in f718712.** These optional params were accessed via guarded `if "x" in inputs and inputs["x"]:`, so no KeyError was actually possible, but I converted them to `inputs.get("x")` to match the SDK contract (required → `inputs["x"]`, optional → `inputs.get("x")`). Now **0**. Semantics are unchanged, including the falsy `convert_lead` booleans where the `in inputs` membership guard is retained.

2. **`defined in input_schema but never accessed` (288) — false positives, left as-is.** The checker only walks each action's `execute()` body and does not trace into helper functions. All these params are consumed inside the `build_*_data` / `build_*_query_params` / `build_search_params` helpers (e.g. `create_contact` builds its payload via `build_contact_data(inputs)`). Eliminating them would mean inlining 12 helpers across ~50 actions — large, risky churn that would worsen the code, so I've left the helper architecture intact.

Tram-Nguyen87:
@TheRealAgentK — ready for review. Summary of what's been done and **fresh** test runs on the current head (`83304bf`).

## What this PR is
SDK 1.0.2 → 2.0.0 migration for Zoho CRM (all 54 actions), kept tightly scoped to the migration plus one review fix:

- **SDK 2.0 response handling** — every `context.fetch()` call site reads `FetchResponse.data`; error paths converted to `ActionError(message=...)`, including Zoho soft errors (non-`SUCCESS` code, empty envelope).
- **Codex P2 fix (`83304bf`)** — Zoho returns HTTP 204 / no body for empty modules, which SDK 2 surfaces as `response.data is None`. Added a `response_body()` helper (`response.data or {}`) routed through all ~90 read sites, so an empty module returns `[]` / a clean not-found instead of crashing into a misleading `ActionError`. +26 regression tests covering the `None`-body path (list/search/related/COQL → empty; single-record get → graceful not-found).

## Fresh test runs (current head, not stale)
| Suite | Result |
|-------|--------|
| Unit (`pytest zoho/`) | ✅ **235 passed** |
| **Live e2e — read-only** (list ×7, COQL, get-by-id, missing-record) | ✅ **10 passed, 1 expected skip** |
| **Live e2e — destructive** (Contact + Deal create/update/note/delete lifecycles) | ✅ **2 passed** |
| Structure / Code / README / Version (CI) | ✅ all pass |

Live runs were executed against a real Zoho CRM sandbox (AU datacenter, `zohoapis.com.au`) on commit `83304bf`. Destructive lifecycles self-clean (delete steps included and passing) — no residual records.

## Notes for review
- **Code check "⚠️ warnings"** are the pre-existing `defined in input_schema but never accessed` false positives (params consumed inside the `build_*` helpers the checker can't trace) — detailed in the comment above; non-blocking, present on `master` too.
- **Out of scope (flagged for transparency):** `convert_lead` has a pre-existing success-detection bug unrelated to this migration (its `code == "SUCCESS"` check doesn't match Zoho's convert response shape). Deliberately **not** touched here to keep the PR migration-only — happy to raise it as a separate follow-up.

zoho/tests/test_zoho_integration.py
```diff
@@ -0,0 +1,208 @@
...
+_spec = importlib.util.spec_from_file_location("zoho_integration_mod", os.path.join(_parent, "zoho.py"))
+zoho_mod = importlib.util.module_from_spec(_spec)
+_spec.loader.exec_module(zoho_mod)
+
+zoho = zoho_mod.zoho
+
+pytestmark = pytest.mark.integration
+
+ACCESS_TOKEN = os.environ.get("ZOHO_ACCESS_TOKEN", "")
+API_DOMAIN = os.environ.get("ZOHO_API_DOMAIN", "https://www.zohoapis.com")
```
TheRealAgentK, line 39: Please add `ZOHO_ACCESS_TOKEN` and `ZOHO_API_DOMAIN` to the root `.env.example` so the live test setup is discoverable.

zoho/tests/test_zoho_integration.py
```diff
@@ -0,0 +1,208 @@
...
+class TestContactLifecycle:
+    async def test_create_update_note_delete(self, live_context):
+        ts = int(time.time())
+        # create
+        create = await zoho.execute_action(
+            "create_contact",
+            {"Last_Name": "ZohoIT", "First_Name": "Temp", "Email": f"zoho.it.{ts}@example.com"},
+            live_context,
+        )
+        contact_id = _assert_ok(create)["contact"]["id"]
```
TheRealAgentK, line 157: Please wrap the destructive lifecycle cleanup in `finally` once IDs are created, so a failed update/assertion doesn’t leave test CRM records behind.

TheRealAgentK (changes requested):
Looks otherwise good. After these changes, please freshly re-run the Zoho integration tests and post the results and then we should be good.

Tram-Nguyen87:
@TheRealAgentK addressed both review comments and freshly re-ran the suite. 

**Changes (`799045e`)**
- **`finally` cleanup for destructive tests** — once IDs are created, the delete/cleanup now runs in a `finally` block in both `TestContactLifecycle` (contact + note) and `TestDealLifecycle`, so a failed update/assertion no longer leaves records behind in the live CRM.
- **`.env.example`** — added a `Zoho CRM` section documenting `ZOHO_ACCESS_TOKEN` and `ZOHO_API_DOMAIN` (with the datacenter-domain note).
- The earlier Codex P2 (treat 204/empty body as empty result) was already addressed in `83304bf` via the `response_body()` helper applied across all list/search/related-record actions.

**Fresh live run** (read-only + destructive, against the live CRM):
```
pytest zoho/tests/test_zoho_integration.py -m "integration" -v
...
12 passed, 1 skipped in 6.85s
```
The single skip is `TestSearchReadOnly::test_search_contacts_word` — Zoho returns 204 on a no-match search, which the test treats as a clean skip. The full create/update/note/delete lifecycle (including the new `finally` cleanup path) passed end-to-end.

**Unit suite** also green: `235 passed`.

TheRealAgentK:
Fixed the conflict, approved and merged.

TheRealAgentK (approved):

PR #375, opened 2026-06-17
feat(webcal): upgrade to SDK 2.0.0 with unit + integration tests

Tram-Nguyen87:
## Summary

Upgrades the **Webcal** integration from `autohive-integrations-sdk` **1.0.2 → 2.0.0**.

SDK 2.0.0's only breaking change is that `context.fetch()` now returns a `FetchResponse` (`.status`, `.headers`, `.data`) instead of the raw body. Webcal has a single fetch call site, in `WebCalendarAPI.fetch_calendar()`, so the migration is small and self-contained. Webcal has **no error-return paths**, so no `ActionError` conversion was required, and it uses **no authentication** (it fetches public iCal/webcal feeds).

## Changes

- **`webcal.py`** — read the iCal body from `response.data` before parsing with `icalendar` (the one breaking-change site).
- **`requirements.txt`** — pin `autohive-integrations-sdk~=2.0.0`.
- **`config.json`** — bump version `1.0.0 → 2.0.0` (major bump for the breaking SDK dependency).
- **`__init__.py`** — re-export the integration (`from .webcal import webcal`), matching the calendly/x/bigquery convention and resolving the package-import collision under pytest's importlib mode.
- **Tests** — add a full pytest suite and remove the stale pre-migration `test_webcal.py` / `context.py`:
  - `tests/test_webcal_unit.py` — **24 unit tests**, mocked `context.fetch` returning `FetchResponse`. Covers both actions across every branch: time-range / ongoing / out-of-range filtering, all-day (with and without an end date), recurring flag, full + single-attendee + organizer + url metadata, missing-`dtend` fallback, timezone handling, `webcal://`→`https://` rewrite, search matching across summary/description/location with precedence, case sensitivity, empty/no-result cases, and `ResultType.VALIDATION_ERROR` for both actions' missing required inputs.
  - `tests/test_webcal_integration.py` — **7 integration tests** using a `live_context` fixture whose `real_fetch` returns `FetchResponse` with the body read as **text** (iCal is `text/calendar`, not JSON). Hits a public US-Holidays feed; overridable via `WEBCAL_TEST_URL`.
  - `tests/conftest.py` — adds the integration dir to `sys.path`.

## Testing

| Check | Result |
|---|---|
| Unit tests | ✅ 24 passed |
| Integration tests (live feed) | ✅ 7 passed |
| `validate_integration.py` | ✅ 0 errors, 0 warnings |
| `check_code.py` | ✅ lint, format, bandit, pip-audit, config-sync, fetch-pattern |

Integration tests never run in CI (the default `-m unit` filter excludes them and they must be invoked by explicit path), so the external public feed cannot break the pipeline.


webcal/tests/test_webcal_integration.py
```diff
@@ -0,0 +1,134 @@
...
+sys.path.insert(0, _parent)
+
+import webcal as webcal_mod  # noqa: E402
+
+webcal_integration = webcal_mod.webcal
+
+pytestmark = pytest.mark.integration
+
+# Public US Holidays iCal feed — no auth required.
+PUBLIC_ICS_URL = os.environ.get("WEBCAL_TEST_URL", "https://www.calendarlabs.com/ical-calendar/ics/76/US_Holidays.ics")
```
TheRealAgentK, line 32: Please add `WEBCAL_TEST_URL` to the root `.env.example` so the optional live-test override is documented.

TheRealAgentK (changes requested):
Code looks good. No need to re-run integration tests for this docs-only fix; once `WEBCAL_TEST_URL` is added to `.env.example`, this should be ready to approve.

Tram-Nguyen87:
Done — added a `Webcal` section to the root `.env.example` documenting the optional `WEBCAL_TEST_URL` override (commit 526c041). Docs-only change, no test re-run needed.

TheRealAgentK (approved):

PR #377, opened 2026-06-17
fix(.env.example): sort integration sections alphabetically to avoid merge conflicts

Shubhank-Jonnada:
Sorts all `# -- Section --` blocks in `.env.example` alphabetically so that future PRs adding new integration env vars don't cascade into conflicts at the bottom of the file.


Shubhank-Jonnada:
No longer needed — master already has the alphabetical sorting applied. Feature branches have been fixed to only add their own sections in the correct alphabetical position.

PR #384, opened 2026-06-22
feat(substack): upgrade to SDK 2.0.0 with unit + integration tests

Tram-Nguyen87:
## Summary

Migrates the **Substack** integration to `autohive-integrations-sdk~=2.0.0`.

Substack's public API requires no authentication, so this integration has no credentials, no auth config, and no error-signalling action returns — it relies on `context.fetch` raising `HTTPError`/`RateLimitError` on non-2xx (the SDK v2 behaviour).

### Source (`substack.py`)
- **FetchResponse unwrap** — access the body via `.data` at all 5 `context.fetch` call sites: `get_publication_posts`, `get_post`, `search_publications`, `search_posts`, `get_post_comments`. The two dict-guarded sites assign `body = response.data` first so `isinstance(...)` checks the body, not the `FetchResponse`.
- **No `ActionError` conversion** — the integration has no `ActionResult(data={"error": ...})` patterns; the 1.x source had no `try/except` either, so the no-error-handling stance is unchanged and consistent. Hardening is out of pure-migration scope.

### `config.json`
- Version `1.0.0` → `2.0.0`.
- **Comment item fields made nullable** (`body`, `date`, `author_name`, `author_id`, `like_count`, `id`). Comments are returned as raw API passthrough (unlike posts/publications, which `_drop_none` cleans). The live API returns `body: null` on deleted comments — without nullable types, SDK v2 output validation recurses into the array items and raises `VALIDATION_ERROR` on a successful 200. **Confirmed against the live API**, not speculative.

### `requirements.txt`
- Pin `autohive-integrations-sdk~=2.0.0`.

### Tests
- Removed the legacy `unittest` suite (`tests/context.py`, `tests/test_substack.py`).
- **Unit (`test_substack_unit.py`) — 38 tests**: every action, `FetchResponse` mocks, `_drop_none` null-field handling, validation errors, and a regression test (`test_null_comment_fields_pass_output_validation`) reproducing the null-body case the schema fix addresses.
- **Integration (`test_substack_integration.py`) — 14 tests, live, no auth**: every action against a real public publication (default `https://www.astralcodexten.com`, overridable via `SUBSTACK_TEST_PUBLICATION_URL`), plus a bad-slug test asserting `HTTPError` propagates so the `real_fetch` raise path is actually exercised. `real_fetch` returns `FetchResponse` and raises `HTTPError`/`RateLimitError` on non-ok per repo convention.

## Verification (run on the final commit, `efbbf02`)

| Check | Result |
|---|---|
| `validate_integration.py` | ✅ 0 errors, 0 warnings |
| `check_code.py` (lint, format, bandit, pip-audit, config-sync, fetch patterns) | ✅ CODE CHECK PASSED |
| Unit tests | ✅ 38 passed |
| Integration tests (live) | ✅ 14 passed, 1 skipped |

## Self-review

Reviewed against the `upgrading-sdk-v2`, `writing-unit-tests`, and `writing-integration-tests` skills, plus an independent fresh-context review of the branch diff. The comment-nullability bug and the missing error-path test coverage were both **found and fixed pre-emptively** during that review rather than left for a reviewer to flag. The 1 skip is expected (`get_post_comments` skips when a post has no numeric id).


Tram-Nguyen87:
confirmed and fixed in 4c8360f.

`tests/conftest.py` was prepending the `substack/` directory to `sys.path`, which makes `substack.py` importable as a top-level module and shadows the `substack` package. Reproduced in a clean interpreter: with `substack/` on `sys.path[0]`, `import substack` resolves to `substack/substack.py` (a module, not a package), so the package-style imports (`from substack import substack`, `from substack.substack import _normalise_url`) fail at collection.

Fix: removed the `sys.path` manipulation and rely on repo-root package resolution (the root `conftest.py` puts the repo root on `sys.path`, and `--import-mode=importlib` does not re-insert test dirs). This matches the `circle` integration, which uses the same `__init__` re-export + package imports with no path hack. The conftest is kept (the validator requires `tests/conftest.py`) but now only documents the import strategy.

Verified after the fix: unit 38 passed, integration 14 passed / 1 skipped (live), `validate_integration.py` 0 errors, `check_code.py` passed.

TheRealAgentK (changes requested):
Requesting one small should-fix before merge:

`substack/tests/test_substack_integration.py` reads `SUBSTACK_TEST_PUBLICATION_URL` and `SUBSTACK_TEST_POST_SLUG`, but this PR does not add blank template entries for them in the root `.env.example`. Even though both are optional, the integration-test checklist asks that every env var read by integration tests is documented there.

Suggested addition:

```env
# -- Substack --
# Optional: publication URL used by Substack live integration tests.
# SUBSTACK_TEST_PUBLICATION_URL=
# Optional: post slug used by Substack get_post live integration tests.
# SUBSTACK_TEST_POST_SLUG=
```

Everything else in the SDK v2 migration, tests, and validation looked good.

TheRealAgentK (approved):
LGTM now

PR #387, opened 2026-06-23
feat(monday-com): migrate to autohive-integrations-sdk 2.0.0

Tram-Nguyen87:
## Summary

Migrates the Monday.com integration from `autohive-integrations-sdk~=1.0.2` to `~=2.0.0`.

**Source changes (`monday_com.py`)**
- `execute_graphql_query`: wraps `context.fetch()` result via `FetchResponse.data`; added None-body guard to surface empty responses as a clean `ActionError` rather than a confusing `TypeError`
- All 6 action handlers: converted every error return from `ActionResult(data={\"error\": ...})` to `ActionError(message=...)` per SDK 2.0.0 contract
- `get_users`: removed `teams` sub-field from the GraphQL query — confirmed via live API testing that the field requires elevated scopes beyond `users:read` and returns `UNAUTHORIZED_FIELD_OR_TYPE`

**Config changes (`config.json`)**
- Bumped integration version `0.1.0` → `2.0.0`
- Removed `error` and `result` fields from all 6 action output schemas (now signalled via `ActionError` return type)

**Tests**
- Replaced old `unittest`-based `test_monday_com.py` + `context.py` with a full pytest suite
- `test_monday_com_unit.py`: 35 unit tests covering all 6 actions — happy path, request shape, GraphQL error → `ACTION_ERROR`, exception → `ACTION_ERROR`, missing required inputs → `VALIDATION_ERROR`
- `test_monday_com_integration.py`: live API tests for `get_boards`, `get_items`, `get_users`; destructive `create_item → update_item → create_update` lifecycle (opt-in via `-m destructive`)

## Test plan

- [x] 35 unit tests passing — `pytest monday-com/tests/test_monday_com_unit.py -v`
- [x] 9 integration tests passing against live Monday.com API — `pytest monday-com/tests/test_monday_com_integration.py -m integration -v`
- [x] `validate_integration.py` — 0 errors (1 pre-existing warning: `updates:write` scope)
- [x] `check_code.py` — all checks green (syntax, imports, lint, format, bandit, pip-audit, config-code sync, fetch patterns)
- [ ] Destructive `create_item` lifecycle — skipped (requires `MONDAY_TEST_BOARD_ID` env var pointing to a test board)

Screenshot 2026-06-23 142922


monday-com/tests/test_monday_com_integration.py
```diff
@@ -0,0 +1,284 @@
...
+
+from monday_com import monday_com as integration
+
+pytestmark = pytest.mark.integration
+
+ACCESS_TOKEN = os.environ.get("MONDAY_ACCESS_TOKEN", "")
+
+# Optional: provide a known board/item so tests don't need to chain list→get.
+TEST_BOARD_ID = os.environ.get("MONDAY_TEST_BOARD_ID", "")
+TEST_ITEM_ID = os.environ.get("MONDAY_TEST_ITEM_ID", "")
```
TheRealAgentK: 🚫 Blocker: these new live integration tests read `MONDAY_ACCESS_TOKEN`, `MONDAY_TEST_BOARD_ID`, and `MONDAY_TEST_ITEM_ID`, but the PR doesn't add matching blank entries to the root `.env.example`. Per the integration-test checklist, every env var read by integration tests must be documented there so reviewers and future contributors have a canonical setup contract. Please add the variables that remain in this test file, or remove any unused env-var reads.

monday-com/tests/test_monday_com_integration.py
```diff
@@ -0,0 +1,284 @@
...
+    assert update_result.result.data.get("item") is not None
+
+    # Comment
+    comment_result = await integration.execute_action(
+        "create_update",
+        {"item_id": item_id, "body": "Autohive integration test comment — automated"},
+        live_context,
+    )
+    assert comment_result.type == ResultType.ACTION, comment_result.result.message
+    assert comment_result.result.data.get("update") is not None
```
TheRealAgentK: ⚠️ Should-fix: this destructive lifecycle creates and comments on a real Monday.com item, but there is no cleanup. If Monday supports archiving/deleting the item via GraphQL, please wrap the lifecycle in a `try/finally` and clean up the created test item. If cleanup is not supported, explicitly document that this test leaves data behind and should only run against a disposable test board.

monday-com/tests/test_monday_com_integration.py
```diff
@@ -0,0 +1,284 @@
...
+async def _first_item_id(live_context, board_id):
+    """Return an item id from a board, or skip if none exist."""
+    if TEST_ITEM_ID:
+        return TEST_ITEM_ID
+    from autohive_integrations_sdk import ResultType
+
+    result = await integration.execute_action("get_items", {"board_id": board_id, "limit": 1}, live_context)
+    if result.type != ResultType.ACTION:
+        pytest.skip(f"get_items failed: {result.result.message}")
+    items = result.result.data.get("items", [])
```
TheRealAgentK: ⚠️ Should-fix: `_first_item_id()` and its `MONDAY_TEST_ITEM_ID` env var appear to be unused by the integration tests. Either add coverage that uses this helper, or remove both to avoid unnecessary setup/documentation surface.

TheRealAgentK (commented):
Posted the blocker and should-fix integration-test comments inline.

TheRealAgentK:
⚠️ Should-fix: `validate_integration.py monday-com` reports `updates:write` as a potentially unused scope. It looks likely intentional because `create_update` should need it, but please either resolve the validator warning if possible or acknowledge why the scope is required before merge.

TheRealAgentK (changes requested):
Requesting changes because the new live integration tests read Monday.com env vars that are not documented in root `.env.example`, which is required by the integration-test checklist.

TheRealAgentK (approved):
Approved. Previous review issues have been addressed; remaining note about PR-body command wording is non-blocking.

PR #390, opened 2026-06-24
feat(heygen): upgrade to SDK 2.0.0 with unit + integration tests

Tram-Nguyen87:
## Summary

- Upgrades HeyGen integration from \`autohive-integrations-sdk~=1.0.2\` to \`~=2.0.0\`
- Converts all 19 action handlers to the SDK 2.0 \`FetchResponse\` pattern (\`response.data\`)
- Replaces all error returns with \`ActionError\`
- Fixes a \`list_voice_locales\` output schema bug that caused \`VALIDATION_ERROR\` on successful API responses (locale item fields were non-nullable but the real API returns \`null\` for some)
- Replaces the old script-style \`test_heygen.py\` with a full pytest suite: 70 unit tests + integration tests
- Clears all config-code sync warnings by replacing \`inputs[\"key\"]\` with \`inputs.get(\"key\")\` for every optional parameter
- Fixes \`sys.path\` in \`conftest.py\` and test files to insert the repo root so \`import heygen.heygen\` resolves correctly as a package submodule in all environments

### Source (\`heygen.py\`)
- All \`context.fetch()\` results access \`.data\` for the response body
- All \`except Exception\` handlers return \`ActionError(message=str(e))\` instead of \`ActionResult\` with error data
- \`list_avatars\` simplification logic updated to use \`response.data\` correctly
- All optional parameter accesses use \`inputs.get(\"key\")\` to clear config-code sync warnings
- Added \`ActionError\` to SDK import

### Config (\`config.json\`)
- \`list_voice_locales\` output schema: locale item fields (\`locale\`, \`language\`, \`region\`) changed from \`\"string\"\` to \`[\"string\", \"null\"]\` — real API returns null for some fields, causing VALIDATION_ERROR on 200 responses

### Requirements
- \`autohive-integrations-sdk~=1.0.2\` → \`~=2.0.0\`

### Tests
- Replaced \`tests/context.py\` + \`tests/test_heygen.py\` with:
  - \`tests/conftest.py\` — inserts repo root on \`sys.path\` so \`heygen\` resolves as the package
  - \`tests/test_heygen_unit.py\` — 70 unit tests covering all 19 actions (happy paths, \`ActionError\` error paths, \`ResultType.VALIDATION_ERROR\`, optional field handling, URL construction, \`list_avatars\` simplification logic, nullable locale fields)
  - \`tests/test_heygen_integration.py\` — live API tests for all read-only actions; \`@pytest.mark.destructive\` class for credit-costing operations (video creation, avatar group lifecycle)
- All fetch mocks use \`FetchResponse(status=200, headers={}, data=...)\`
- \`real_fetch\` in integration tests returns \`FetchResponse\`

### Docs / Env
- \`README.md\`: updated testing section, added missing action docs (\`list_avatars_in_group\`, \`get_photo_avatar_details\`), added SDK 2.0 version history entry, fixed broken \`USAGE_EXAMPLE.md\` reference
- \`.env.example\`: added \`HEYGEN_ACCESS_TOKEN\` and \`HEYGEN_TEST_IMAGE_KEY\` (optional, for destructive tests)

## Test plan

- [x] \`validate_integration.py heygen\` — ✅ passed
- [x] \`check_code.py heygen\` — ✅ passed, zero warnings (lint, format, bandit, pip-audit, config-code sync, fetch patterns)
- [x] \`run_tests.py heygen\` — ✅ 70/70 passed
- [x] \`check_readme.py\` — ✅ passed
- [x] \`check_version_bump.py\` — ✅ passed
- [x] \`hiveup package heygen\` — ✅ built successfully
- [x] Beta agent test: \`list_voices\` ✅, \`list_voice_locales\` ✅ (was failing before schema fix), \`get_video_status\` ✅

## Notes

Remaining beta failures are environment/account issues, not code:
- **401 on photo avatar actions** — connected HeyGen API key lacks photo avatar / account-level permissions; needs full-access key from HeyGen dashboard
- **403 on video creation** — free-tier OAuth apps cannot generate videos; requires paid HeyGen plan
- **\`list_avatars\` timeout** — endpoint returns large payload for accounts with many avatars; simplification logic is already in place but a large account can still hit the 30s Lambda limit


heygen/tests/test_heygen_integration.py
```diff
@@ -0,0 +1,496 @@
...
+        pytest.skip("HEYGEN_ACCESS_TOKEN not set — skipping integration tests")
+
+    async def real_fetch(url, *, method="GET", params=None, headers=None, json=None, **kwargs):
+        async with aiohttp.ClientSession() as session:
+            async with session.request(method, url, params=params, json=json, headers=headers or {}) as resp:
+                try:
+                    resp_data = await resp.json(content_type=None)
+                except Exception:
+                    resp_data = await resp.text()
+                return FetchResponse(status=resp.status, headers=dict(resp.headers), data=resp_data)
```
TheRealAgentK, line 53: Should fix before merge: this live `real_fetch` helper doesn't mirror SDK 2.0.0 behavior for non-2xx responses. The real `ExecutionContext.fetch` parses the body and then raises `HTTPError` when `response.ok` is false, so handlers hit their `except Exception -> ActionError` path. This fixture returns `FetchResponse` for every status, which means the invalid-ID integration tests can pass without exercising the production error path.

Please either make this helper raise on non-2xx like the SDK does, or use the real SDK `ExecutionContext` for those tests, then assert `ResultType.ACTION_ERROR` for non-2xx API failures.

TheRealAgentK (changes requested):
Thanks for the SDK v2 upgrade and the expanded tests. Local validation was green, but I’m requesting changes for one integration-test contract issue: the custom live fetch helper does not mirror SDK 2.0.0 non-2xx behavior, so the invalid-ID tests can give false confidence about the production `ActionError` path.

heygen/tests/test_heygen_integration.py
```diff
@@ -0,0 +1,498 @@
+"""
+End-to-end integration tests for the HeyGen integration.
+
+These tests call the real HeyGen API and require a valid OAuth access token
+set in the HEYGEN_ACCESS_TOKEN environment variable.
+
+Run all read-only tests:
+    pytest heygen/tests/test_heygen_integration.py -m integration
```
TheRealAgentK: Blocker: this command is documented as the read-only integration-test command, but `pytestmark = pytest.mark.integration` applies to the whole module and the destructive classes also carry the `integration` marker. `-m integration` will therefore run the credit-costing/destructive tests too. Please change the read-only command here (and the matching README command) to `-m "integration and not destructive"`, leaving `-m "integration and destructive"` as the explicit opt-in command.

TheRealAgentK (changes requested):
Thanks for fixing the previous HeyGen review issues. The import path issue and SDK-like non-2xx `real_fetch` behavior both look fixed. I’m requesting changes for one remaining integration-test safety blocker: the documented read-only command still runs destructive tests.

heygen/tests/test_heygen_integration.py
Shubhank-Jonnada: fixed in commit [ebb3709](https://github.com/Autohive-AI/autohive-integrations/pull/390/commits/ebb37090901e5c788ae65a821feeca1fa5add7fc). The read-only test command in both the test file docstring and README.md has been updated from -m integration to -m "integration and not destructive".

TheRealAgentK (approved):
Re-review approved. The previous blocker is fixed: the read-only integration-test command now uses `-m "integration and not destructive"` in both the test file and README. I also rechecked structure validation and unit tests locally; both passed.

PR #393, opened 2026-06-26
feat(microsoft-planner): upgrade to autohive-integrations-sdk 2.0.0

Tram-Nguyen87:
## Summary

- Migrates Microsoft Planner from `autohive-integrations-sdk~=1.0.2` to `~=2.0.0`
- All 35 action handlers updated: `context.fetch()` results now access `.data` for the response body
- All error returns converted from `ActionResult(data={"error": ...})` to `ActionError(message=...)`
- `"error"` fields removed from all 28 action output schemas in `config.json`
- `import uuid` moved to module level (was incorrectly inside `AddChecklistItemAction.execute`)
- `config.json` version bumped `1.0.0` → `2.0.0`

## Key implementation notes

**ETag pattern**: Microsoft Planner requires a current ETag (`If-Match` header) for every PATCH and DELETE. A `get_etag()` helper fetches the resource first and extracts `@odata.etag`. The helper catches all exceptions internally and returns `None`, so ETag-failure returns `ActionError("Failed to retrieve X ETag")` before the mutating request.

**PATCH 204 No Content**: All Planner PATCH endpoints return 204 with no body. Affected handlers guard with `response.data if response.data is not None else {}`.

**Checklist actions**: `add_checklist_item`, `update_checklist_item`, `remove_checklist_item` fetch task details directly (not via `get_etag()`). The fetch result variable is named `details_response` to avoid the CI fetch-pattern linter false-positive on `response.data.get(...)`.

**Platform OAuth**: Auth shape is `{"auth_type": "PlatformOauth2", "credentials": {"access_token": "..."}}`. The SDK injects the bearer token automatically; no custom auth header logic needed in handlers.

## Tests

Replaced the old SDK 1.x `test_microsoft_planner.py` (unittest-based, bare dict mocks) with:

| File | Domain | Tests |
|---|---|---|
| `test_microsoft_planner_users_unit.py` | list_groups, get_user_by_email, search_users, get_current_user, list_user_tasks, list_user_plans | 29 |
| `test_microsoft_planner_plans_unit.py` | list_plans, get_plan, create_plan, update_plan, delete_plan, get_plan_details, update_plan_details | 33 |
| `test_microsoft_planner_buckets_unit.py` | list_buckets, get_bucket, create_bucket, update_bucket, delete_bucket, list_bucket_tasks | 28 |
| `test_microsoft_planner_tasks_unit.py` | list_tasks, get_task, create_task, update_task, delete_task, get_task_details, update_task_details, add/update/remove_checklist_item | 40 |
| `test_microsoft_planner_board_formats_unit.py` | 6 board format actions (get + update × 3 types) | 21 |
| **Total** | **All 35 actions** | **141 tests** |

Integration test file covers all 35 actions: all read-only actions as individual test classes, plus lifecycle tests for plans, buckets, tasks, checklists, and board formats.

## Test plan

**Run unit tests (CI equivalent):**
```bash
python -m pytest microsoft-planner/tests/test_microsoft_planner_*_unit.py -v
# Expected: 141 passed
```

**Run integration tests — read-only (safe, requires access token):**
```bash
python -m pytest microsoft-planner/tests/test_microsoft_planner_integration.py -m "integration and not destructive" -v
```

**Run destructive integration tests (requires access token + test resource IDs):**
```bash
python -m pytest microsoft-planner/tests/test_microsoft_planner_integration.py -m "integration and destructive" -v
```

Required env vars (see root `.env.example`):
- `MICROSOFT_PLANNER_ACCESS_TOKEN` — required for any integration test
- `MICROSOFT_PLANNER_TEST_GROUP_ID` — optional, enables list_plans + plan lifecycle tests
- `MICROSOFT_PLANNER_TEST_PLAN_ID` — optional, enables bucket/task read tests
- `MICROSOFT_PLANNER_TEST_BUCKET_ID` — optional, enables task lifecycle tests
- `MICROSOFT_PLANNER_TEST_TASK_ID` — optional, enables task detail + board format tests
- `MICROSOFT_PLANNER_TEST_USER_EMAIL` — optional, enables get_user_by_email + search_users tests

## Self-review against skills

Applied: `upgrading-sdk-v2`, `writing-unit-tests`, `writing-integration-tests`, `reviewing-integration-prs`

- [x] All `context.fetch()` results access `.data`
- [x] All error paths return `ActionError(message=...)`
- [x] `"error"` fields removed from all output schemas
- [x] `requirements.txt` pins `~=2.0.0`; `config.json` version bumped to `2.0.0`
- [x] Unit test mocks use `FetchResponse(status, headers, data)`
- [x] Error assertions use `result.type == ResultType.ACTION_ERROR`
- [x] Integration test `live_context` (Variant 3 Platform OAuth) returns `FetchResponse`
- [x] All 35 actions covered in unit tests; all 35 covered in integration tests
- [x] All integration test env vars documented in root `.env.example`
- [x] `ruff check` and `ruff format --config tooling/ruff.toml` clean
- [x] `validate_integration.py` passed (1 pre-existing warning: `offline_access` scope)
- [x] 141/141 unit tests pass (fresh run post all changes)


microsoft-planner/microsoft_planner.py
```diff
@@ -191,11 +176,11 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext) -> Ac
 
             response = await context.fetch(f"{GRAPH_API_BASE_URL}/users/{user_id}/planner/tasks", method="GET")
```
TheRealAgentK: Blocker: when `user_id` defaults to `"me"`, this builds `/users/me/planner/tasks`, but Microsoft Graph's current-user alias is a top-level segment: `/me/planner/tasks`. The same issue exists in `list_user_plans` with `/users/me/planner/plans`; please branch on `user_id == "me"` and use `/me/...` for the default current-user case, while keeping `/users/{id}/...` for actual user IDs. The unit tests should be updated too, since they currently lock in the invalid `/users/me/...` URL.

microsoft-planner/tests/test_microsoft_planner_integration.py
```diff
@@ -0,0 +1,622 @@
...
+        pytest.skip("MICROSOFT_PLANNER_ACCESS_TOKEN not set — skipping integration tests")
+
+    import aiohttp
+
+    async def real_fetch(url, *, method="GET", json=None, headers=None, params=None, **kwargs):
+        merged_headers = dict(headers or {})
+        merged_headers["Authorization"] = f"Bearer {token}"
+        async with aiohttp.ClientSession() as session:
+            async with session.request(method, url, json=json, headers=merged_headers, params=params) as resp:
+                data = await resp.json(content_type=None)
```
TheRealAgentK: Blocker: this live fetch helper always parses JSON, but Planner update/delete endpoints commonly return `204 No Content`. That can raise after a real mutation succeeds, making destructive lifecycle tests report an `ActionError` or obscure cleanup state. Please handle `204` / empty bodies before JSON parsing and return `FetchResponse(status=resp.status, headers=dict(resp.headers), data=None)` for no-content responses.

microsoft-planner/tests/test_microsoft_planner_integration.py
```diff
@@ -0,0 +1,622 @@
...
+@pytest.mark.destructive
+class TestBucketLifecycle:
+    """End-to-end workflow: create bucket → update → delete."""
+
+    @pytest.mark.asyncio
+    async def test_full_bucket_lifecycle(self, live_context):
+        require_plan_id()
+
+        # Step 1: Create
+        create_result = await microsoft_planner.execute_action(
```
TheRealAgentK, line 406: Should-fix: the destructive lifecycle tests create real Planner resources but do not use `try/finally` cleanup. If create succeeds and a later update/details/assertion fails, the bucket/task/plan can be left behind in the tenant. Please track created IDs and attempt best-effort cleanup in `finally` blocks for the bucket, task/checklist, and plan lifecycle tests.

TheRealAgentK (changes requested):
Thanks for the Microsoft Planner SDK v2 upgrade and the expanded tests. I’m requesting changes for the blockers called out inline: the default current-user endpoints use `/users/me/...` instead of `/me/...`, and the live integration-test fetch helper does not handle Planner `204 No Content` responses. I also left one should-fix around cleanup safety for destructive lifecycle tests.

microsoft-planner/microsoft_planner.py
Shubhank-Jonnada: Fixed in commit f231e2e. list_user_tasks and list_user_plans now branch on user_id == 'me' and use /me/planner/tasks and /me/planner/plans for the default current-user case, keeping /users/{id}/... for actual user IDs. Unit test assertions updated to match.

microsoft-planner/tests/test_microsoft_planner_integration.py
Shubhank-Jonnada: Fixed in commit 62c63a3. The real_fetch helper now checks resp.status == 204 or empty content length before attempting JSON parsing, returning FetchResponse(data=None) for no-content responses. Also added non-2xx raising behavior to raise HTTPError or RateLimitError to mirror SDK behavior.

microsoft-planner/tests/test_microsoft_planner_integration.py
Shubhank-Jonnada: Fixed in commit 6a3be14. All four destructive lifecycle tests (bucket, task, checklist, plan) now wrap intermediate steps in try/finally blocks so cleanup runs even if an assertion fails mid-test.

microsoft-planner/tests/test_microsoft_planner_integration.py
```diff
@@ -0,0 +1,636 @@
...
+        pytest.skip("MICROSOFT_PLANNER_ACCESS_TOKEN not set — skipping integration tests")
+
+    import aiohttp
+
+    async def real_fetch(url, *, method="GET", json=None, headers=None, params=None, **kwargs):
+        merged_headers = dict(headers or {})
+        merged_headers["Authorization"] = f"Bearer {token}"
+        async with aiohttp.ClientSession() as session:
+            async with session.request(method, url, json=json, headers=merged_headers, params=params) as resp:
+                if resp.status == 204 or resp.content_length == 0:
```
TheRealAgentK, line 80: Still blocking: this fixes the 204 success case, but the ordering still returns `FetchResponse` for any empty-body response before checking `resp.status == 429` / `not resp.ok`. That means an empty-body 401/403/404/429/500 from Graph would be treated as success in these live tests, which still diverges from SDK `fetch` behavior. Please move the non-2xx checks before the empty-body success return, or make this guard `if resp.ok and (resp.status == 204 or resp.content_length == 0): ...`.

TheRealAgentK (changes requested):
Re-review: the `/me/planner/...` endpoint fix and destructive-test cleanup are in place, but the live `real_fetch` helper still has a contract mismatch: empty-body non-2xx responses can return as successful `FetchResponse`s before the error checks run. I left the narrowed blocker inline.

PR #394, opened 2026-06-26
fix(hubspot): use v3 API for create_contact

sumitramanga:
### Description :memo:
- **Purpose**: Fix the `create_contact` action which was calling the legacy HubSpot v1 endpoint (`/contacts/v1/contact`), causing a `"property cannot be missing or null"` error because v1 expects properties as a list of `{"property", "value"}` objects, not the v3-style `{"properties": {...}}` dict the code builds.
- **Approach**: Switch the endpoint to `/crm/v3/objects/contacts`, consistent with `update_contact` and the rest of the HubSpot integration. Unit tests updated to assert the correct URL.

**Type of change**
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)

**Updates**
:point_right: Replaced legacy v1 endpoint with v3 API URL in `create_contact`
:point_right: Added inline comment explaining why v3 is required over v1
:point_right: Updated unit tests to assert the correct v3 endpoint URL

### Screenshots :camera:
Create contact action is now running successfully
image

### Test plan :test_tube:
Create contact API endpoint now works

- [x] Self-/dev-tested

### DIFF
--- a/hubspot/hubspot.py
+++ b/hubspot/hubspot.py
@@ -906,7 +906,10 @@ class CreateContactActionHandler(ActionHandler):
         :return: Dictionary with a "contact" key mapping to the created contact data.
         """
 
-        url = "https://api.hubapi.com/contacts/v1/contact"
+        # Use v3 API endpoint. The legacy v1 endpoint expects properties as a
+        # list of {"property": ..., "value": ...} objects, whereas v3 accepts
+        # the {"properties": {...}} dict format we build below.
+        url = "https://api.hubapi.com/crm/v3/objects/contacts"
 
         # Merge properties and additional_properties
         properties = inputs.get("properties", {}).copy()
--- a/hubspot/hubspot.py
+++ b/hubspot/hubspot.py
@@ -906,9 +906,6 @@ class CreateContactActionHandler(ActionHandler):
         :return: Dictionary with a "contact" key mapping to the created contact data.
         """
 
-        # Use v3 API endpoint. The legacy v1 endpoint expects properties as a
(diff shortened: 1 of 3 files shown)


sumitramanga:
Re the structure warning - it reckons some scopes aren't in use but they seem like they are. This would be a false positive

sumitramanga:
I can look into the code warning post this work. We actually need to do some migration work that HubSpot is asking about. So I can bucket it in there

hubspot/hubspot.py
```diff
@@ -906,7 +906,10 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
         :return: Dictionary with a "contact" key mapping to the created contact data.
         """
 
-        url = "https://api.hubapi.com/contacts/v1/contact"
+        # Use v3 API endpoint. The legacy v1 endpoint expects properties as a
+        # list of {"property": ..., "value": ...} objects, whereas v3 accepts
+        # the {"properties": {...}} dict format we build below.
```
ProRedCat: We don't need this comment

sumitramanga: I can remove, I thought to keep for context as this was the only endpoint that uses v1 API

sumitramanga:
```suggestion
```

ProRedCat: Generally comments like these aren't helpful as they're mentioning past behaviour that doesn't matter anymore. In this case v1 HubSpot endpoints are very old and should have been on v3 like the rest of the integration

ProRedCat (approved):

PR #395, opened 2026-06-26
feat: Add ProjectWorks integration

sumitramanga:
### Description :memo:
- **Purpose**: Adds a new ProjectWorks integration to the autohive-integrations repo, providing full read and write access to the core ProjectWorks PSA entities (users, clients, projects, modules, tasks, resources, timesheets, leave, invoices, expense claims, and offices).
- **Approach**: Implements the integration using the `autohive-integrations-sdk`, with HTTP Basic auth (Consumer Key/Secret), a shared set of helper functions (`_list`, `_get`, `_write`, `_delete`) to reduce boilerplate, and a dedicated expense-claim file-attachment handler that supports both base64 content and URL-based file downloads.

**Type of change**

**Updates**
:point_right: Added `projectworks/projectworks.py` with 40+ actions covering read (`list_*`, `get_*`) and write (`create_*`, `update_*`, `delete_*`) operations across all core ProjectWorks entities, plus sub-resource actions for project/task assignments, user roles, leave balances, postings, and custom fields
:point_right: Added `projectworks/config.json` defining the integration schema, auth configuration (HTTP Basic with Consumer Key/Secret), and all action input/output definitions
:point_right: Added unit and integration test suites (`test_projectworks_unit.py`, `test_projectworks_integration.py`) with mocked and live-API coverage, plus a destructive timesheet lifecycle test; updated `.env.example` with ProjectWorks credential variables

### Screenshots :camera:
beta carmack dev_w-wry6s_chat_a_1336616_1429578 (1)

### Test plan :test_tube:
1. Run unit tests (no credentials needed): `pytest projectworks/`
2. Set `PROJECTWORKS_CONSUMER_KEY` and `PROJECTWORKS_CONSUMER_SECRET` in `.env`
3. Run read-only integration tests: `pytest projectworks/tests/test_projectworks_integration.py -m "integration and not destructive"`
4. Optionally run destructive tests (creates/updates/deletes a real timesheet): `pytest projectworks/tests/test_projectworks_integration.py -m "integration and destructive"`

- [x] Self-/dev-tested
- [x] Appropriate documentation written (code comments, internal docs)


sumitramanga:
PR is not ready just yet - I needed to trigger the validation tests to see what issues to fix up

sumitramanga:
Ran the skill for reviewing integration PR's. Here's the results. I'll gitignore the DS_store but I'm unsure about the other 2 suggestions  - 
image

sumitramanga:
Integration test results using the consumer key and secret on a test account - 
image

sumitramanga:
I tested the integraiton within the app too. All tests pass in the end - https://beta.carmack.dev/shared/fytx7ajtk8wphge7vpjkvth8bsno5ulk

One didn't initially pass because we didn't supply it an ID

Another didn't pass because it had already been cascade-deleted when its parent project (ID 10) was deleted.

Running both on their own worked perfectly

sumitramanga:
@TheRealAgentK Can I get your recomendation on the 2 suggestions the review skill got back to me with please

projectworks/config.json
```diff
@@ -0,0 +1,3038 @@
...
+          "help_text": "From ProjectWorks -> Admin -> API Accounts -> create an account -> Details. Used as the Basic auth username."
+        },
+        "consumer_secret": {
+          "type": "string",
+          "format": "password",
+          "label": "Consumer Secret",
+          "help_text": "Shown only once when the API account is created. Used as the Basic auth password."
+        }
+      },
+      "required": []
```
dre-raygun, line 26: We should be making both of these fields required I think.

dre-raygun (changes requested):
Only found one blocking thing, and although I'm happy to not block the PR on it, something to consider though is perhaps following a 'search' leaning action set instead of 'listing'. Would be better for the context windows of consuming LLMs.

Otherwise, I've pulled the integration down, packaged + uploaded and tested it on my local and everything works as advertised 🙌

projectworks/projectworks.py
```diff
@@ -0,0 +1,1251 @@
...
+    them in whatever form the target API expects.
+    """
+    content_b64 = file_obj.get("content")
+    if content_b64:
+        try:
+            return base64.b64decode(content_b64)
+        except Exception:
+            raise ValueError("file 'content' is not valid base64-encoded data")
+
+    url = file_obj.get("url")
```
TheRealAgentK: Should-fix: this `file.url` path makes the integration download an arbitrary caller-provided URL from the runtime, then upload those bytes to ProjectWorks. That creates SSRF and large-response/memory risk. Could we remove URL download support for now and require the standard file object with base64 `content` instead?

projectworks/tests/test_projectworks_integration.py
```diff
@@ -0,0 +1,292 @@
...
+        data = ok_data(result)
+        assert isinstance(data["offices"], list)
+
+
+# ---- Destructive Tests (Write Operations) ----
+# These create, update, and delete real data on the connected account.
+# Only run deliberately with: pytest -m "integration and destructive"
+
+
+@pytest.mark.destructive
```
TheRealAgentK, line 308: Should-fix: could we add a little broader integration-test coverage/evidence before merge? The current live tests are good smoke coverage plus the timesheet lifecycle, but this PR adds a lot of write/sub-resource actions. Even a few more representative live tests or documented evidence for the higher-risk action families would make this easier to approve confidently.

TheRealAgentK (commented):
Thanks for the ProjectWorks integration — leaving two should-fix notes inline.

projectworks/config.json
TheRealAgentK: No, please don't. There's currently an investigation into the behaviour of `required` and auth with @NinosMan and @ProRedCat happening and there's potentially something fishy going on if required is used on auth fields. I'll post a separate comment on this on your PR, @sumitramanga

projectworks/projectworks.py
```diff
@@ -0,0 +1,1251 @@
...
+                if resp.status != 200:
+                    raise ValueError(f"Failed to download file from url: HTTP {resp.status}")
+                return await resp.read()
+
+    raise ValueError("file object missing 'content' (base64) or 'url' — attach a receipt file to the expense claim")
+
+
+def _get_headers(context: ExecutionContext) -> Dict[str, str]:
+    creds = context.auth.get("credentials", context.auth)
+    consumer_key = creds.get("consumer_key", "")
```
TheRealAgentK, line 30: Small auth note: I don’t think we should blindly add `auth.fields.required` here unless we confirm the exact custom-auth shape the platform passes at runtime. If production passes wrapped credentials, required flat fields could cause SDK validation to fail before the handler runs.

Could we instead add a small runtime guard in `_get_headers()` so missing/blank Consumer Key or Secret returns a clear error rather than sending `Basic Og==`? That keeps the schema compatible while avoiding confusing unauthenticated requests.

```python
if not consumer_key or not consumer_secret:
    raise ValueError("ProjectWorks Consumer Key and Consumer Secret are required")
```

projectworks/projectworks.py
TheRealAgentK: Security should-fix: the `file.url` option makes this action download an arbitrary caller-provided URL from the integration runtime and then upload those bytes to ProjectWorks. That creates SSRF risk, and `resp.read()` also allows large responses to be buffered and base64-expanded in memory.

Could we remove URL download support for now and require the standard Autohive file object with base64 `content`? That seems like the smallest safe path for this PR.

```python
content_b64 = file_obj.get("content")
if not content_b64:
    raise ValueError("file object missing 'content' (base64)")
```

TheRealAgentK (commented):
Adding the auth and file-download notes inline with suggested code snippets.

projectworks/projectworks.py
TheRealAgentK: This is related to my response to @dre-raygun's comment above.

TheRealAgentK (changes requested):

projectworks/projectworks.py
TheRealAgentK: A bit more detail on why I think removing `url` is the safest fix here: this is different from passing a public URL through to a third-party API. The integration runtime itself is making the request, so a caller-controlled URL can point at internal/private targets such as localhost, private RFC1918 addresses, or cloud metadata/link-local addresses. Redirects are also followed by default unless explicitly disabled/rechecked.

The 25s timeout avoids hangs, but it does not limit response size. `await resp.read()` buffers the whole response in memory, then the code base64-encodes it, which expands the payload before sending it to ProjectWorks. So this is both SSRF risk and large-response/memory risk.

Unless we have a platform-approved safe downloader, I’d prefer to keep expense receipts to the standard Autohive file object with base64 `content` only.

TheRealAgentK:
> @TheRealAgentK Can I get your recomendation on the 2 suggestions the review skill got back to me with please

First one is really a stylistic thing. If you want to get it out, I'd not bother at this stage and fix in a later addition/ticket.

Second one is correct, I think, that's what I posted too.

I got Amp to explain me the attack/sec vector in more detail.

projectworks/config.json
sumitramanga: I initially had this but as Kai said, something fishy is going on there 🙂  It seems like it was breaking the integration actions

TheRealAgentK (approved):
This looks good now! Thx for the changes, @sumitramanga

dre-raygun (approved):

PR #398, opened 2026-07-01
feat: Projectworks rename casing, add search actions, field projection, and default page size

sumitramanga:
### Description :memo:
- **Purpose**: Renames the integration from "ProjectWorks" to "Projectworks" (correct brand casing), adds free-text search actions for users, clients, and projects, introduces optional field projection on all `list_*` actions, and applies a conservative default page size to prevent unbounded list responses.
- **Approach**: A `_project()` helper trims list records to caller-specified fields. A `_search()` helper scans one page client-side and matches a query substring against relevant fields (with composed-field support for full-name matching). `DEFAULT_PAGE_SIZE` (50) is applied when no `page_size` is provided; `DEFAULT_SEARCH_SCAN` (200) is used for search scans. All changes are covered by new unit and integration tests.

**Type of change**
- [ ] Bug fix (non-breaking change which fixes an issue)

**Updates**
:point_right: Renamed all references from "ProjectWorks" to "Projectworks" across config, source, README, and tests; bumped version to `1.1.0`
:point_right: Added `search_users`, `search_clients`, and `search_projects` actions — free-text, case-insensitive substring search over a single scanned page with `limit` and `fields` support
:point_right: Added optional `fields` projection to all `list_*` actions and applied `DEFAULT_PAGE_SIZE = 50` so unparameterised calls stay bounded

### Screenshots :camera:
{Image here of before and after - if applicable}

### Test plan :test_tube:
*Provide guidance for how to QA your proposed changes. This is not only for a test but also useful for a reviewer.*

- Run unit tests: `pytest -m unit` — covers `_project`, `_matches_query`, default page size enforcement, field projection on `list_users`, and all three search actions including limit capping, composed-field matching, and validation errors.
- Run integration tests (requires valid `.env` credentials): `pytest -m integration` — covers `TestListFieldProjection` (projection + default page size bounds) and `TestSearchUsers/Clients/Projects` against the live API.
- Manually verify the renamed display name appears correctly in the Autohive UI when the integration is loaded.

- [x] Self-/dev-tested

### DIFF
--- a/projectworks/config.json
+++ b/projectworks/config.json
@@ -1,6 +1,6 @@
 {
   "name": "Projectworks",
-  "version": "1.0.1",
+  "version": "1.1.0",
   "description": "Projectworks integration for reading users, clients, projects, modules, tasks, resourcing, timesheets, leave, invoices, and expense claims from your professional services automation account.",
   "display_name": "Projectworks",
   "supports_connected_account": false,
@@ -64,8 +64,15 @@
           },
           "page_size": {
             "type": "integer",
-            "description": "Results per page (default 100)",
+            "description": "Results per page (default 50)",
             "minimum": 1
+          },
+          "fields": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Optional list of response field names (PascalCase, e.g. [\"UserID\", \"FirstName\", \"Email\"]) to return for each record. When omitted, full records are returned. Use this to keep large result sets small for an LLM context window, then call the matching get_* action for the full record."
           }
         },
         "required": []
@@ -128,8 +135,15 @@
           },
           "page_size": {
             "type": "integer",
-            "description": "Results per page (default 100)",
+            "description": "Results per page (default 50)",
             "minimum": 1
+          },
+          "fields": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Optional list of response field names (PascalCase, e.g. [\"UserID\", \"FirstName\", \"Email\"]) to return for each record. When omitted, full records are returned. Use this to keep large result sets small for an LLM context window, then call the matching get_* action for the full record."
           }
         },
         "required": []
@@ -184,8 +198,15 @@
           },
           "page_size": {
             "type": "integer",
-            "description": "Results per page (default 100)",
+            "description": "Results per page (default 50)",
             "minimum": 1
+          },
+          "fields": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Optional list of response field names (PascalCase, e.g. [\"UserID\", \"FirstName\", \"Email\"]) to return for each record. When omitted, full records are returned. Use this to keep large result sets small for an LLM context window, then call the matching get_* action for the full record."
           }
         },
         "required": []
@@ -276,8 +297,15 @@
           },
           "page_size": {
             "type": "integer",
-            "description": "Results per page (default 100)",
+            "description": "Results per page (default 50)",
             "minimum": 1
+          },
+          "fields": {
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "description": "Optional list of response field names (PascalCase, e.g. [\"UserID\", \"FirstName\", \"Email\"]) to return for each record. When omitted, full records are returned. Use this to keep large result sets small for an LLM context window, then call the matching get_* action for the full record."
           }
         },
         "required": []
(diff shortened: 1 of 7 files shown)


sumitramanga:
Tests are passing and manual tests in the app are looking good. I have the latest package up on beta

Regarding the version bump, is it a major one appropriate to do, considering what the validator discovered?

projectworks/config.json
```diff
@@ -1067,6 +1158,156 @@
...
       }
     },
+    "search_users": {
+      "display_name": "Search Users",
+      "description": "Free-text search for users by name or email. Scans one page (default 200 records) and returns up to `limit` lean matches. A convenience layer over list_users for \"find the user called ...\" flows; for exhaustive enumeration use list_users with explicit filters and pagination.",
+      "input_schema": {
+        "type": "object",
+        "properties": {
+          "query": {
+            "type": "string",
```
TheRealAgentK, line 1168: This new search action accepts blank queries because `query` is only constrained as a string. With the current substring matching, `query: ""` will match every string field (`"" in "Jane"` is true), so `search_users` returns the first page of users instead of rejecting an invalid search.

Can we make `query` non-blank for all three search actions, e.g. add `minLength: 1` plus `pattern: "\\S"` here and in `search_clients` / `search_projects`? I’d also consider trimming/rejecting blank queries in `_search()` as a defensive runtime guard.

projectworks/tests/test_projectworks_integration.py
```diff
@@ -300,6 +300,79 @@ async def test_returns_offices(self, live_context):
...
+            pytest.skip("No users in account to search")
+        target = listed[0]
+        term = target.get("FirstName") or target.get("Name") or target.get("Email")
+        if not term:
+            pytest.skip("Listed user has no name/email field to search on")
+
+        result = await projectworks.execute_action("search_users", {"query": term, "fields": ["UserID"]}, live_context)
+        users = ok_data(result)["users"]
+        assert isinstance(users, list)
+        assert all(set(u).issubset({"UserID"}) for u in users)
```
TheRealAgentK, line 344: This derives a query from a real listed user, but it doesn’t assert that the searched user is actually returned. If `search_users` returned `[]`, this test would still pass because `all(...)` over an empty list is true.

Can we assert the target ID appears in the results? For example:

```python
target_id = target.get("UserID")
...
assert any(u.get("UserID") == target_id for u in users)
```

Same idea for the client and project search tests below: since each query is derived from a listed page-1 record, the test should verify that record’s ID is present in the search results.

projectworks/projectworks.py
```diff
@@ -145,6 +179,69 @@ async def _delete(
     return ActionResult(data={deleted_key: deleted_value, "deleted": True}, cost_usd=0.0)
 
 
+def _matches_query(
+    record: Dict[str, Any],
+    query: str,
+    search_fields: List[str],
+    composed_fields: List[List[str]] | None = None,
+) -> bool:
+    q = query.casefold()
```
TheRealAgentK, line 188: Because `query` is not normalized or checked before `_matches_query()`, blank input turns search into an arbitrary first-page list. In Python, `"" in value.casefold()` is always true, and whitespace can also match composed names like `"Jane Doe"`.

Small defensive fix:

```python
query = query.strip()
if not query:
    raise ValueError("query must not be blank")
```

near the start of `_search()`, plus schema validation for the three `query` fields.

TheRealAgentK:
`fields` is accepted as any array, including `[]`, but `_project()` treats empty lists as omitted and returns full records. That’s surprising for a projection parameter and weakens the context-size safety goal if a caller/UI accidentally sends an empty array.

Can we either reject empty arrays with `minItems: 1` (and preferably `items.minLength: 1`) on the `fields` schemas, or explicitly document that omitted-or-empty means “return full records”? I’d lean toward rejecting empty arrays.

TheRealAgentK:
> Tests are passing and manual tests in the app are looking good. I have the latest package up on beta
> 
> Regarding the version bump, is it a major one appropriate to do, considering what the validator discovered?

Can do either, I'd personally be fine with 1.1.0 as you did it because I can't see a breaking change. Sometimes the validator is a bit over-hyped to go major update.

sumitramanga:
> > Tests are passing and manual tests in the app are looking good. I have the latest package up on beta
> > Regarding the version bump, is it a major one appropriate to do, considering what the validator discovered?
> 
> Can do either, I'd personally be fine with 1.1.0 as you did it because I can't see a breaking change. Sometimes the validator is a bit over-hyped to go major update.

I'll leave it as is, thank you!

TheRealAgentK (approved):
LGTM now - we've discussed the edge case around `fields` if it's an empty array and decided to leave as-is for now.

PR #408, opened 2026-07-03
feat(gitlab): migrate to SDK 2.0.0

Shubhank-Jonnada:
## Summary
- Upgrades the GitLab integration from `autohive-integrations-sdk~=1.0.2` to `~=2.0.0`
- All `context.fetch()` call sites updated to read the body via `.data` (breaking change in 2.0.0)
- All error paths converted from `ActionResult(data={"error": ...})` to `ActionError(message=...)`
- Rewrote the test suite: mocked unit tests (`test_gitlab_unit.py`, 53 tests covering all 20 actions) and a live-API integration suite (`test_gitlab_integration.py`) using the `live_context`/`FetchResponse` pattern
- Added `GITLAB_ACCESS_TOKEN` / `GITLAB_PROJECT_ID` to root `.env.example`

## Test plan
- [x] `python -m pytest gitlab/tests/test_gitlab_unit.py -v` — 53/53 passed
- [x] `validate_integration.py gitlab` — all checks passed
- [x] `check_code.py gitlab` — passed (config-code sync warnings are pre-existing false positives from loop-based param extraction, not related to this migration)
- [ ] Live integration tests — pending a GitLab OAuth token/test project (see note to reviewer)


Shubhank-Jonnada:
## Live API verification

All 20 actions in this integration were tested against the real GitLab API using a live personal access token (scopes: `read_api`, `read_user`, `read_repository`, `read_registry`), run against the public `gitlab-org/gitlab-foss` project (no owned projects were available on the test account).

| Category | Actions | Result |
|---|---|---|
| User | `get_current_user` | ✅ |
| Projects | `list_projects`, `get_project` | ✅ |
| Issues | `list_issues`, `get_issue` | ✅ |
| Merge Requests | `list_merge_requests`, `get_merge_request`, `get_merge_request_changes`, `list_merge_request_commits` | ✅ |
| Branches | `list_branches`, `get_branch` | ✅ |
| Commits | `list_commits`, `get_commit`, `get_commit_diff` | ✅ |
| Pipelines | `list_pipelines`, `get_pipeline`, `list_pipeline_jobs` | ✅ |
| Repository | `list_repository_tree`, `get_file`, `get_file_raw`, `compare_branches` | ✅ |
| Container Registry | `list_container_registry_repositories`, `get_container_registry_repository`, `list_container_registry_tags`, `get_container_registry_tag` | ✅ |

Live testing also caught a real bug (fixed in 59b4b6c): several boolean query params (`owned`, `membership`, `starred`, `all`, `with_stats`, `first_parent`, `include_retried`, `recursive`, `tags`, `tags_count`) were passed as raw Python `bool` values, which `aiohttp` rejects in query strings (only `str`/`int`/`float` allowed). All are now stringified to `"true"`/`"false"`, matching the pattern already used elsewhere in the file (e.g. `statistics`, `straight`).

The token used for testing has since been rotated.

gitlab/tests/test_gitlab_integration.py
```diff
@@ -0,0 +1,203 @@
...
+        merged_headers["Authorization"] = f"Bearer {ACCESS_TOKEN}"
+        async with aiohttp.ClientSession() as session:
+            async with session.request(
+                method, url, json=json, data=body, headers=merged_headers, params=params
+            ) as resp:
+                try:
+                    data = await resp.json(content_type=None)
+                except Exception:
+                    data = await resp.text()
+                return FetchResponse(status=resp.status, headers=dict(resp.headers), data=data)
```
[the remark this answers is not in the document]
Shubhank-Jonnada: fixed in 48434de. The \`real_fetch\` shim in \`test_gitlab_integration.py\` now raises \`HTTPError\`/\`RateLimitError\` on non-2xx responses (mirroring \`ExecutionContext.fetch()\`'s real behavior), instead of always returning a \`FetchResponse\`. Previously a 401/404 body would have quietly looked like an empty successful list rather than exercising the \`ActionError\` path. This now matches the same pattern already used in dropbox/circle/calendly's integration test shims.

Shubhank-Jonnada:
@codex review

gitlab/config.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "GitLab",
   "display_name": "GitLab",
-  "version": "1.0.0",
+  "version": "2.0.0",
```
TheRealAgentK, line 4: `gitlab/README.md` still documents only `1.0.0` in its Version History even though this changes the integration version to `2.0.0`. Please add a `2.0.0` entry covering the SDK v2 migration / `FetchResponse.data` handling / `ActionError` failure contract.

gitlab/requirements.txt
```diff
@@ -1 +1 @@
-autohive-integrations-sdk~=1.0.2
+autohive-integrations-sdk~=2.0.0
```
TheRealAgentK: Current validation warns that `autohive-integrations-sdk~=2.0.0` is deprecated and recommends `~=2.0.1` or later. Please bump this pin or explicitly confirm we want to keep the deprecated range for this migration.

gitlab/config.json
```diff
@@ -31,10 +31,6 @@
           "user": {
             "type": "object",
             "description": "User information including id, username, email, name, avatar_url"
-          },
```
TheRealAgentK, line 34: The schema removal is correct, but `gitlab/README.md` still documents this old in-band response contract (`result`, optional `error`) and repeats `result: Success status` in each action output section. Please update the README so it describes SDK v2 behavior: successful actions return only action-specific data, and failures surface as `ActionError` / action errors rather than response fields.

TheRealAgentK (changes requested):
Requesting changes for the inline comments already left.

gitlab/config.json
Shubhank-Jonnada: Fixed in 652a440 (docs(gitlab): add 2.0.0 entry to README version history), added a 2.0.0 entry to the Version History covering the SDK v2 migration.

gitlab/requirements.txt
Shubhank-Jonnada: Fixed in 652a440 (fix(gitlab): bump SDK pin to ~=2.0.1), bumped the pin to clear the deprecation warning.

gitlab/config.json
Shubhank-Jonnada: Fixed in 652a440 (docs(gitlab): document SDK v2 action result contract), rewrote the Action Results section and removed the stale `result: Success status` line from every action's Outputs, documenting that success returns action-specific data only and failures surface as `ActionError`.

TheRealAgentK (approved):

PR #409, opened 2026-07-03
feat(eventbrite): migrate to SDK 2.0.0

Shubhank-Jonnada:
## Summary
- Upgrades the Eventbrite integration from `autohive-integrations-sdk~=1.0.2` to `~=2.0.0`
- All `context.fetch()` call sites updated to read the body via `.data` (breaking change in 2.0.0)
- All error paths — both `except` blocks and manual validation guards (e.g. missing `event_id`) — converted from `ActionResult(data={"result": False, "error": ...})` to `ActionError(message=...)`
- Switched required-field extraction from `inputs.get("x")` to `inputs["x"]` to match the schema's `required` list and keep `check_config_sync.py` fully clean (zero warnings)
- Rewrote the test suite: 80 mocked unit tests covering all 27 actions (including validation-error and guard-error paths) and a live-API integration suite with a gated destructive create/update/delete event lifecycle test
- Added `EVENTBRITE_ACCESS_TOKEN` / `EVENTBRITE_ORGANIZATION_ID` / `EVENTBRITE_RUN_DESTRUCTIVE_TESTS` to root `.env.example`

## Test plan
- [x] `python -m pytest eventbrite/tests/test_eventbrite_unit.py -v` — 80/80 passed
- [x] `validate_integration.py eventbrite` — all checks passed, 0 warnings
- [x] `check_code.py eventbrite` — passed, 0 warnings (config-code sync fully clean)
- [ ] Live integration tests — pending an Eventbrite OAuth token/org ID (see note to reviewer)


Shubhank-Jonnada:
## Live API verification

All 27 actions in this integration were tested against the real Eventbrite API using a live personal token, against the "Autohive Engineering" test organization.

**Read actions (16):** get_current_user, list_organizations, list_categories, get_category, list_events, get_event, get_event_description, list_venues, get_venue, list_orders_by_event, list_orders_by_organization, list_attendees, list_ticket_classes, get_ticket_class — all ✅. (`get_order` and `get_attendee` were skipped since no real orders/attendees exist on the test event — expected for a sandbox event with no signups.)

**Write actions (11):** create_event, update_event, publish_event, unpublish_event, copy_event, create_venue, create_ticket_class, update_ticket_class, delete_ticket_class, cancel_event, delete_event — all ✅. Note: `publish_event` initially failed with a real Eventbrite validation error (`event.tickets - MISSING`) because the event had no ticket class yet — confirmed this is correct API behavior (Eventbrite requires a ticket class before publishing), not a bug, by retrying with a ticket class present, which then published cleanly. Test events/venues/ticket classes were cleaned up (deleted/cancelled) after the run.

Also proactively applied the same fix flagged by the codex bot on the gitlab PR: the `real_fetch` shim in `test_eventbrite_integration.py` now raises `HTTPError`/`RateLimitError` on non-2xx responses (e2dced4), matching `ExecutionContext.fetch()`'s real behavior instead of silently returning an empty success on auth/API errors.

The token used for testing has since been rotated.

eventbrite/tests/test_eventbrite_integration.py
```diff
@@ -0,0 +1,154 @@
...
+        merged_headers["Authorization"] = f"Bearer {ACCESS_TOKEN}"
+        async with aiohttp.ClientSession() as session:
+            async with session.request(
+                method, url, json=json, data=body, headers=merged_headers, params=params
+            ) as resp:
+                try:
+                    data = await resp.json(content_type=None)
+                except Exception:
+                    data = await resp.text()
+                return FetchResponse(status=resp.status, headers=dict(resp.headers), data=data)
```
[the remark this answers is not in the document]
Shubhank-Jonnada: This was addressed in [e2dced4 ](https://github.com/Autohive-AI/autohive-integrations/pull/409/commits/e2dced47dbbacb0a14ef0fbe21fe5ec3a7be1dfc). The real_fetch shim now raises HTTPError/RateLimitError on non-2xx responses instead of always returning a FetchResponse, so it correctly mirrors ExecutionContext.fetch() and exercises the ActionError path on auth/API failures rather than silently returning an empty success.

Shubhank-Jonnada:
@codex review

eventbrite/config.json
```diff
@@ -33,17 +33,9 @@
                     "user": {
                         "type": "object",
                         "description": "User details"
-                    },
-                    "result": {
-                        "type": "boolean",
```
TheRealAgentK, line 38: The schema removal is correct, but `eventbrite/README.md` still documents this old in-band response contract (`result` / `error`) and repeats `result` in the action output sections, including an error-handling block showing `{ "result": false, "error": ... }`. Please update the README to match SDK v2 behavior: successful actions return action-specific data only, and failures surface as `ActionError` / action errors rather than response fields.

eventbrite/eventbrite.py
```diff
@@ -112,33 +117,30 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
             response = await context.fetch(url, method="GET")
 
             return ActionResult(
-                data={"result": True, "events": response.get("events", []), "pagination": response.get("pagination")},
+                data={
+                    "events": response.data.get("events", []),
+                    "pagination": response.data.get("pagination"),
```
TheRealAgentK: This can return `pagination: None` if Eventbrite omits the key, but the output schema declares `pagination` as `type: object`. SDK v2 validates output data, so that would turn a successful API response into a validation error. Please only include `pagination` when it is present, or explicitly allow `null` in the schema. Same pattern appears in the other list actions returning pagination.

eventbrite/tests/test_eventbrite_integration.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+# ---- Destructive event lifecycle ----
+
+
+class TestEventLifecycle:
+    @skip_if_not_destructive
+    @skip_if_no_org
+    @pytest.mark.asyncio
+    async def test_01_create_update_delete_event(self, live_context):
```
TheRealAgentK, line 141: This destructive lifecycle is env-gated, but it should also be marked `@pytest.mark.destructive`. It also only deletes the created event at the final step; if create succeeds and update/assert fails, the event is left behind. Please wrap cleanup in a `try/finally` and mark the test destructive.

eventbrite/requirements.txt
```diff
@@ -1 +1 @@
-autohive-integrations-sdk~=1.0.2
+autohive-integrations-sdk~=2.0.0
```
TheRealAgentK: Current validation warns that `autohive-integrations-sdk~=2.0.0` is deprecated and recommends `~=2.0.1` or later. Please bump this pin or explicitly confirm we want to keep the deprecated range for this migration.

TheRealAgentK (changes requested):
Requesting changes for the inline comments already left.

eventbrite/config.json
Shubhank-Jonnada: Fixed in d458800 (docs(eventbrite): document SDK v2 action result contract), removed the stale `result`/`error` lines from every action's Outputs and rewrote the Error Handling section to describe the SDK v2 contract (data-only success, `ActionError` on failure).

eventbrite/eventbrite.py
Shubhank-Jonnada: Fixed in 016c7ad (fix(eventbrite): omit pagination key when absent to satisfy output schema), `pagination` is now only included in the response when the API actually returns it, so it no longer conflicts with the `type: object` schema. Applied to all 7 affected actions.

eventbrite/tests/test_eventbrite_integration.py
Shubhank-Jonnada: Fixed in 23955cd (test(eventbrite): mark event lifecycle test destructive and guarantee cleanup), added `@pytest.mark.destructive` and wrapped the update/assert step in try/finally so the created event is always deleted, even if update fails.

eventbrite/requirements.txt
Shubhank-Jonnada: Fixed in 49743ba (fix(eventbrite): bump SDK pin to ~=2.0.1), bumped the pin to clear the deprecation warning.

TheRealAgentK (dismissed):
LGTM apart from conflicts that need fixing.

TheRealAgentK (approved):

PR #410, opened 2026-07-03
feat(microsoft-word): migrate to SDK 2.0.0

Shubhank-Jonnada:
## Summary
- Upgrades the Microsoft Word integration from `autohive-integrations-sdk~=1.0.2` to `~=2.0.0`
- All `context.fetch()` call sites updated to read the body via `.data` (breaking change in 2.0.0), including binary docx/PDF content downloads that previously relied on the fetch result *being* raw bytes
- All error paths — both `except` blocks and manual validation guards — converted from `ActionResult(data={"result": False, "error": ...})` to `ActionError(message=...)`
- Fixed `word_list_documents` returning `next_page_token: null` on the last page, which fails SDK 2.0's stricter output-schema validation (`"type": "string"` doesn't accept `null`) — now the key is simply omitted when there's no next page
- Removed the dead `include_formatting` param from `word_get_tables`'s input schema — it was never read in code (docstring already said "not implemented")
- Rewrote the test suite: 33 mocked unit tests covering all 9 actions (including validation-error and guard-error paths, binary-content edge cases) and a live-API integration suite with a gated destructive document lifecycle test
- Added `MICROSOFT_WORD_ACCESS_TOKEN` / `MICROSOFT_WORD_RUN_DESTRUCTIVE_TESTS` to root `.env.example`

## Test plan
- [x] `python -m pytest microsoft-word/tests/test_microsoft_word_unit.py -v` — 33/33 passed
- [x] `validate_integration.py microsoft-word` — passes (1 pre-existing benign warning: "potentially unused scopes" — Files.Read/Files.ReadWrite are used by every Graph API call, it's a static-analysis limitation)
- [x] `check_code.py microsoft-word` — passed, 0 warnings (config-code sync fully clean)
- [ ] Live integration tests — pending a Microsoft Graph OAuth token (see note to reviewer)


TheRealAgentK:
Future of this integration is currently under evaluation, NFA for the time being.

PR #414, opened 2026-07-08
docs(rss-reader-feedparser): mark legacy integration unsupported

TheRealAgentK:
## Summary

Marks the legacy `rss-reader-feedparser` integration as unsupported while keeping the directory in place for reference.

`rss-reader-feedparser-ah-fetch` is now the recommended RSS reader integration because it routes HTTP requests through the Autohive SDK `context.fetch` layer and supports basic/Bearer auth.

## Changes

- Added a Heartbeat-style unsupported notice to `rss-reader-feedparser/README.md`
- Updated the root integration catalog to label `rss-reader-feedparser` as unsupported
- Pointed users to `rss-reader-feedparser-ah-fetch` as the maintained replacement

## Validation

- [x] `check_readme.py origin/master rss-reader-feedparser`
- [x] `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ This repository hosts Autohive integrations made and maintained by the Autohive 
 
 ### RSS Reader 
 
-[rss-reader-feedparser](rss-reader-feedparser): Reads RSS feeds using the `feedparser` library. Includes 1 action.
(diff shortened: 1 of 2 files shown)


risheetperi (approved):
LGTM

PR #420, opened 2026-07-14
feat: Add Land Information New Zealand (LINZ) integration

sumitramanga:
### Description :memo:
- **Purpose**: Adds a new LINZ (Land Information New Zealand) integration that reads NZ property, title, ownership, and parcel data from the LINZ Data Service (LDS) WFS API.
- **Approach**: Implements a new `linz` integration module using the `autohive-integrations-sdk`, querying the LDS WFS API via HTTP. Auth is handled via a per-user LINZ Data Service API key (custom auth). A paginated scan-and-aggregate pattern powers the headline `find_multi_property_owners` action. XML exception reports from LDS are parsed to surface clear, actionable error messages (including licence hints for the personal data layer).

**Type of change**

**Updates**
:point_right: Adds `linz/` integration module with 6 actions: `list_available_layers`, `find_multi_property_owners`, `search_property_titles`, `get_title_owners`, `search_parcels`, and `query_layer`
:point_right: Adds unit tests (mocked WFS) and integration tests (real LDS API, read-only) with graceful skipping when the key lacks the Personal Data Licence for `layer-50805`
:point_right: Updates repo-level `README.md` and `.env.example` with LINZ integration entry and required env vars

### Screenshots :camera:
Please lmk if you'd like to see the results 

### Test plan :test_tube:
**Unit tests (no network required):**
```bash
python -m pytest linz/tests/test_linz_unit.py -q
```

**Integration tests (requires a LINZ Data Service API key):**
```bash
LINZ_API_KEY=... pytest linz/tests/test_linz_integration.py -m "integration and not destructive"
```
- Ownership layer tests (`layer-50805`) skip automatically if the key lacks the LINZ Licence for Personal Data
- Optional: set `LINZ_TEST_TITLE_NO` to exercise `get_title_owners` on a known title
- Optional: set `LINZ_TEST_LAND_DISTRICT` to scope multi-property-owner tests (defaults to `Otago`)

- [x] Self-/dev-tested
- [ ] Code is written to standards


linz/tests/conftest.py
```diff
@@ -0,0 +1,18 @@
+import os
+import sys
+
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+# Make linz.py importable as a top-level module.
+sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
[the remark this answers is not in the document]
sumitramanga: Using sys.path now

linz/linz.py
```diff
@@ -0,0 +1,746 @@
...
+
+@linz.action("search_property_titles")
+class SearchPropertyTitlesAction(ActionHandler):
+    """Search NZ property titles (with owners) by owner name, title, or district."""
+
+    async def execute(self, inputs: Dict[str, Any], context: ExecutionContext) -> ActionResult:
+        try:
+            clauses: List[str] = []
+            if inputs.get("owner_name"):
+                clauses.append(f"owners ILIKE {_cql_literal('%' + inputs['owner_name'] + '%')}")
```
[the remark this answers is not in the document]
sumitramanga: the review bot is right, this doesn't exist on the layer so there's no point in having it in place. I've removed it

sumitramanga:
Getting Wayne to fix up the icon

linz/tests/test_linz_integration.py
```diff
@@ -0,0 +1,236 @@
+"""
+End-to-end integration tests for the LINZ integration.
+
+These call the real LINZ Data Service WFS API and require a valid LINZ Data
```
sumitramanga, line 4: WFS stands for Web Feature Service. It's an OGC (Open Geospatial Consortium) standard protocol for querying geographic feature data over HTTP 

You send requests like GetCapabilities (list available layers) or GetFeature (fetch records, optionally filtered) and get back structured data such like GeoJSON or XML.

linz/tests/test_linz_integration.py
```diff
@@ -0,0 +1,236 @@
...
+from autohive_integrations_sdk import FetchResponse
+from autohive_integrations_sdk.integration import ResultType
+
+from linz.linz import linz
+
+pytestmark = pytest.mark.integration
+
+# Optional pre-existing resources to exercise single-record reads without
+# chaining. Each test that needs one skips gracefully when it is unset.
+TEST_TITLE_NO = os.environ.get("LINZ_TEST_TITLE_NO", "")
```
sumitramanga, line 35: The data is quite sensitive as the API key holder is bound by the Privacy Act 2020. I've decided against putting sensitive data in here

linz/requirements.txt
```diff
@@ -0,0 +1,2 @@
+autohive-integrations-sdk~=2.0.1
+defusedxml~=0.7.1
```
sumitramanga, line 3: The extra package is there because the LINZ integration parses XML from an external service.

From Claude - 

> The list_available_layers action fetches the WFS GetCapabilities document from LINZ, which comes back as XML, and _parse_capabilities_layers() parses it at linz.py:271 using DefusedET.fromstring() (imported at linz.py:33). defusedxml is a drop-in replacement for the stdlib xml.etree.ElementTree that's hardened against XML-based attacks — entity-expansion bombs ("billion laughs") and external entity injection (XXE). Since the XML is fetched over the network from a third party, parsing it with the raw stdlib parser would get flagged by the security scan (bandit rule B314); defusedxml is the standard fix.

linz/linz.py
```diff
@@ -0,0 +1,744 @@
...
+                props = _properties(feature)
+                title_no = props.get("title_no")
+                title_record = {
+                    "title_no": title_no,
+                    "land_district": props.get("land_district"),
+                    "estate_description": props.get("estate_description"),
+                    "type": props.get("type"),
+                    "status": props.get("status"),
+                }
+                names = _split_owners(props.get("owners"))
```
TheRealAgentK: 🚫 **Potential correctness blocker — please verify the owner source and avoid parsing an aggregated display field if possible.**

I verified that historical `layer-50805` does have an aggregated `owners` text field, so the earlier claim that this field does not exist was not correct. However, LINZ constructs it using `string_agg(DISTINCT owner, ', ' ORDER BY owner)` over unescaped free-text personal/corporate names. A comma inside a real name is therefore indistinguishable from the separator. Splitting on `","` can invent owners and then report false `property_count` values in the headline workflow.

The LINZ data model exposes three relevant shapes:

1. **`layer-50805` — NZ Property Titles Including Owners:** one title/spatial row with a presentation-oriented aggregated `owners` text field. Convenient for display, but not safely normalisable because names are unescaped.
2. **`layer-50806` — NZ Property Title Owners:** a spatial, denormalised dataset with one row per distinct `(owner, title_no)` and duplicated title geometry. This avoids splitting the aggregate string.
3. **`table-51564` — NZ Property Title Owners List:** the normalised owner dataset with `id`, `tte_id`, `title_no`, `owner_type`, `prime_surname`, `prime_other_names`, and `corporate_name`. It relates through **`table-51566` — NZ Property Title Estates List** (`owners.tte_id -> estates.id`) and then to titles by `title_no`.

So `table-50806` in the current config/docs is not the normalised table identifier: 50806 is a **layer**, while the normalised owner table is 51564.

I do not know whether the configured API key has access to all of these datasets. Please check the actual licensed `GetCapabilities`/`DescribeFeatureType` response. If 50806 or 51564 is available, use a row-oriented source for aggregation. If only 50805 is available, please document that constraint and determine a correctness-safe approach rather than treating comma splitting as reliable. A sanitized real response/schema fixture would make the decision verifiable.

linz/linz.py
TheRealAgentK, line 1: 🚫 **Blocker — please confirm the licensed personal-data flow has been reviewed and covered.**

This integration returns identifiable owner names, aggregates people by the titles they own, and sends those results through Autohive execution/model infrastructure. The LINZ Licence for Personal Data 2.3 requires reasonable safeguards, prohibits distribution to parties that have not agreed to the licence, prohibits public indexing and unsolicited direct-marketing use, and requires downstream amendment/deletion handling within five business days.

I assume Sumitra and the organisation have considered and covered this correctly, but because this is licensed personal data and the integration engineering charter treats user-data handling as a security decision, we need explicit confirmation before merge. In particular, please confirm the position for Autohive/model providers/workflow collaborators as downstream recipients, retention and execution history, deletion propagation, indexing, prohibited use, and breach handling. If there is an internal legal/privacy approval or documented operating decision, referencing it in the PR is sufficient; sensitive details do not need to be posted publicly.

linz/linz.py
```diff
@@ -0,0 +1,744 @@
...
+        )
+        features = _extract_features(collection)
+        collected.extend(features)
+
+        if len(features) < page_size:
+            break  # last page
+        start_index += page_size
+        if len(collected) >= max_records:
+            # There may be more matches than we scanned.
+            truncated = True
```
TheRealAgentK: ⚠️ **`truncated` is incorrect when the result count exactly equals the cap.**

A full page reaching `max_records` does not prove another matching record exists. For example, exactly 2,000 matches with a 2,000 cap are currently reported as truncated. The unit test with exactly two 1,000-row pages codifies this false positive.

Please use a reliable numeric `numberMatched` when available, or request/collect one sentinel record beyond the cap and discard it. Add separate exact-cap and cap-plus-one tests so `truncated` only means data was actually omitted.

linz/linz.py
```diff
@@ -0,0 +1,744 @@
...
+    Power-user escape hatch for layers/tables not covered by the typed actions
+    (e.g. street addresses, geodetic marks, the owner-centric ownership tables).
+    """
+
+    async def execute(self, inputs: Dict[str, Any], context: ExecutionContext) -> ActionResult:
+        try:
+            layer = inputs["layer"]  # required by schema
+
+            include_geometry = bool(inputs.get("include_geometry"))
+            limit = inputs.get("limit") or 100
```
TheRealAgentK: 🚫 **Potential safety/scope issue — please review unrestricted `query_layer`.**

This action accepts any layer/table, does not require a filter, and passes `limit` directly to WFS without a runtime or schema maximum. A caller can therefore request `layer-50805` without CQL, ask for an arbitrarily large number of licensed owner records, include large geometries, or query datasets that were never intentionally reviewed as Autohive workflows. That bypasses the 10,000-record safety cap used by `find_multi_property_owners` and creates avoidable privacy, output-size, memory, and timeout risk.

Please consider removing the generic escape hatch in favour of explicit workflow actions. If it needs to remain, enforce a server-side cap regardless of input, add schema minimum/maximum constraints, reject negative `start_index`, require a meaningful filter, and consider blocking or separately gating licensed ownership datasets. The other user-supplied `limit` fields should be bounded as well.

linz/linz.py
```diff
@@ -0,0 +1,744 @@
...
+    if count is not None:
+        params["count"] = count
+    if start_index is not None:
+        params["startIndex"] = start_index
+    if sort_by:
+        params["sortBy"] = sort_by
+    if srs_name:
+        params["srsName"] = srs_name
+
+    response = await context.fetch(url, method="GET", params=params)
```
TheRealAgentK: 🚫 **Blocker — the SDK can log the LINZ API key and owner-name filters.**

LINZ requires the API key in the URL path, and this request passes that credential-bearing URL to `context.fetch()`. The current SDK appends `params` to the URL before making the request, so CQL filters containing searched owner names are also part of that URL. On an unexpected error, SDK 2.0.1 logs the complete method and URL. This can expose both `services;key=<secret>/wfs` and personal owner-name search terms in logs.

There is a second runtime mismatch here: production `context.fetch()` raises `HTTPError` for non-2xx responses, so `_check_wfs_response()` never receives the 400/403 `FetchResponse` that it expects. The custom integration-test fetch currently returns `FetchResponse` for non-2xx responses, which hides this difference. The intended licence/error mapping therefore does not run through the actual SDK path, and the surrounding `ActionError(message=str(e))` handlers can expose raw exception details.

This needs an SDK-level fix because the integration cannot redact a URL after `fetch()` has logged it. **Ninos and Kai will look at an SDK solution for 2.0.2**, covering URL redaction and a safe HTTP-error contract. Once available, please update this integration to use that contract and add a regression test with a sentinel API key/owner name proving neither appears in logs or `ActionError`.

TheRealAgentK (commented):
Integration engineering charter review completed. Mechanical validation is green; see the five targeted comments for the SDK 2.0.2 credential/error-path follow-up, licensed personal-data confirmation, owner-layer modelling, generic query safety, and truncation accuracy.

linz/linz.py
TheRealAgentK: This was an Amp finding in which it also did research the LINZ layer structure. I haven't validated this in person - but it seems to be very convinced here.

linz/linz.py
TheRealAgentK: We've started to have a look into this and we'll need to trace what the flow is here first. This has an additional complication in that LINZ seem to return an XML blob with an error and their XML contains the API key as well. 

So, as a stop-gap, the complete XML (=the full response) might have to be suppressed here.

FYI @NinosMan

linz/linz.py
TheRealAgentK: First time I've seen my agent have serious legal concerns and expressed them :)

linz/linz.py
```diff
@@ -0,0 +1,744 @@
...
+    return _short(xml)
+
+
+def _is_unknown_layer_exception(text: str) -> bool:
+    """True when LDS reports a feature type as unknown (usually = no licence)."""
+    low = text.lower()
+    return "unknown" in low and ("feature type" in low or "typename" in low or "layer-" in low)
+
+
+def _check_wfs_response(response: Any) -> None:
```
TheRealAgentK, line 386: @sumitramanga 

There are a few more concerns here. 

With the way the integration uses the SDK via `.fetch` it subscribes to the built-in error handling, returning `ActionResult` and `ActionError` or `HTTPError` types.

But in here we add its own error handling on top of that - in some instances for the same cases (400s) - and we're not sure those might be reached. The choice of `RuntimeError` is interesting - would it make more sense to use any of the SDK error types?

It seems that every actual action handler is then in a global (for it) try/catch and there'll be `ActionError` types wrapped around everything.

Could this maybe all be simplified by this integration NOT using the SDK fetch mechanism but custom aiohttp calls? My thinking here would be that you then have full control over what gets exposed (URLs with API keys) and can filter it all out yourself on the integration level? And this would avoid having to patch the SDK then, too.

We have precedence cases for integrations going down that path (custom aiohttp) in case of "weird" situations similar to this.

TheRealAgentK: cc @NinosMan

linz/linz.py
sumitramanga: Have let Kai know in person - The key holder is responsible for how the data is used, hence why we won't be sharing our own key

We have mechanisms in place to avoid data leaking like workspaces to separate private tasks and the standards security practises the platform has

linz/linz.py
sumitramanga: I got Claude to validate this by looking at external resources and had it apply changes.

linz/linz.py
```diff
@@ -0,0 +1,984 @@
...
+            # re-filtering is needed — the CQL filter already matched the
+            # per-row owner field.
+            owners_index: Dict[str, Dict[str, Any]] = {}
+            for feature in scan["features"]:
+                props = _properties(feature)
+                name = (props.get("owner") or "").strip()
+                title_no = props.get("title_no")
+                if not name or title_no is None:
+                    continue
+                key = _owner_key(name)
```
TheRealAgentK, line 911: This groups records solely by normalized display name. `layer-50806` does not provide a stable cross-title person/entity identifier, so unrelated people with the same name can be merged and name variants can split one owner. Please document that this action finds matching owner names across titles, not verified common ownership, and that results require further verification.

TheRealAgentK: We could say something like:

“Find matching owner names across titles.”:

- Matching names are not proof of common identity.
- Same-name individuals may be conflated.
- Name variants may split one person into multiple groups.
- Results are candidates for further verification, not definitive ownership analysis.

linz/linz.py
TheRealAgentK: I think the API key issue is properly fixed. Question is: do you think we need to do anything about owner names potentially leaking into logs and error messages?

In particularly owner names embedded in CQL filters might be leaking into:

- Transport exception messages that include the request URL/query string.
- LINZ/GeoServer XML or JSON errors that echo the submitted CQL (unless you're not exposing any of that XML anymore)
- The resulting user-visible `ActionError`.

TheRealAgentK: Suggestion from Amp on how to improve/fix it:

> - Do not include arbitrary ClientError text in ActionError; use a fixed transport-error message.
> - Do not return raw WFS/XML/JSON error details that may echo CQL. Map known cases to curated messages and use a generic provider-error message otherwise.
> - Add tests asserting that both a sentinel API key and sentinel owner name are absent from:
>   - Transport errors
>   - XML exception reports
>   - Non-2xx JSON responses
>   - Final ActionError

linz/linz.py
```diff
@@ -0,0 +1,984 @@
...
+
+            try:
+                rows_collection = await _wfs_get_features(
+                    context,
+                    TABLE_TITLE_OWNERS_LIST,
+                    cql_filter=f"title_no = {_cql_literal(title_no)}",
+                    count=DEFAULT_PAGE_SIZE,
+                )
+                owner_rows = _extract_features(rows_collection)
+            except RuntimeError:
```
TheRealAgentK: This catches every `RuntimeError`, so timeouts, 5xx responses, malformed responses, and programming failures are all treated as “the normalized table is inaccessible” and silently fall back to inexact parsing. Please catch only the recognized access/unknown-feature response; other runtime failures should return an error.

linz/linz.py
```diff
@@ -0,0 +1,984 @@
...
+
+            min_properties = int(inputs.get("min_properties") or 2)
+            if min_properties < 1:
+                min_properties = 1
+            max_scan = int(inputs.get("max_titles_scanned") or 2000)
+            include_title_details = inputs.get("include_title_details")
+            if include_title_details is None:
+                include_title_details = True
+
+            scan = await _wfs_collect(
```
TheRealAgentK, line 889: Suggestion: these `startIndex` pages are unsorted, so provider ordering is not guaranteed to remain stable across pages or the truncation probe. Consider including a stable row ID and passing `sortBy=id` for deterministic paging.

linz/linz.py
```diff
@@ -0,0 +1,984 @@
...
+    timeout = aiohttp.ClientTimeout(total=WFS_TIMEOUT_SECONDS)
+
+    # Reuse the context's session if present (as context.fetch does); otherwise
+    # create one and cache it on the context so paged scans keep the connection.
+    session = getattr(context, "_session", None)
+    if not isinstance(session, aiohttp.ClientSession):
+        session = aiohttp.ClientSession()
+        context._session = session
+
+    try:
```
TheRealAgentK, line 331: Because this bypasses `context.fetch`, there are currently no automatic retries, exponential backoff, or `Retry-After`/rate-limit semantics; a 429 is returned as a generic WFS error. Please document that limitation in the README for this initial version.

TheRealAgentK (approved):
Looks good now, thanks so much for the progressive improvements! 👍

PR #422, opened 2026-07-14
style(rss-reader): icon

risheetperi:
Icon added to RSS-Reader


NinosMan (approved):

NinosMan:
Version check does not pass since there was no bump, however this is just an icon update so I've approved it regardless.

PR #424, opened 2026-07-29
feat(linkedin-ads): add lead gen and reach metrics to get_ad_analytics

Shubhank-Jonnada:
## Summary

`get_ad_analytics` requested a hardcoded four-field projection, so Lead Gen form metrics and reach were never returned. LinkedIn's `adAnalytics` endpoint only returns metrics explicitly named in the `fields` parameter, so those columns were absent from every response rather than blocked by permissions.

This adds them behind two optional inputs, bumps the API version so bookings can be requested at all, and documents the reporting constraints that come with reach.

### Lead Gen form metrics

`include_leadgen_metrics` (default `true`) adds:

| Field | Meaning |
|-------|---------|
| `oneClickLeadFormOpens` | Form opens |
| `oneClickLeads` | Form submissions |
| `qualifiedLeads` | Qualified leads shared by the advertiser |
| `costPerQualifiedLead` | Spend per qualified lead |
| `appointmentsScheduled` | Appointments booked ("bookings" in Campaign Manager) |
| `viralOneClickLeadFormOpens` / `viralOneClickLeads` | Equivalents from organic reshares |

These require no scope beyond `r_ads_reporting`, which is the only permission the [reporting docs](https://learn.microsoft.com/en-us/linkedin/marketing/integrations/ads-reporting/ads-reporting) list for `adAnalytics`. `r_ads_leadgen_automation` applies to the Lead Sync API (`/leadFormResponses`, the submitted lead records), which this integration does not use.

### Reach and frequency

`include_reach` (default `false`) adds `approximateMemberReach` and `audiencePenetration`, plus a derived `frequency` (`impressions / reach`) since LinkedIn has no frequency field.

Opt-in because both metrics are only served for date ranges of 92 days or less. Past that LinkedIn returns `200` with the fields **silently absent** rather than an error, so the range is validated up front and a longer range returns a clear action error instead of quietly incomplete data.

### API version 202601 to 202607

`appointmentsScheduled` was added in `202605`. On `202601` it fails the entire request:

```
400 Projected field "appointmentsScheduled" not present in schema
    "com.linkedin.adsexternalapi.reportingapi.v8.AdAnalyticsV8"
```

No breaking change between `202602` and `202607` affects the endpoints used here — the only one in that window is `metricType` on the member post/video analytics APIs, which this integration does not call.

### Other

- Reject reversed date ranges (`end_date` before `start_date`), which previously passed through to the API
- Guard LinkedIn's documented 20-metric cap on `fields`

## Validation

Read-only verification against a live ad account on `202607`. No write actions were called.

- [x] All seven GET actions pass: `get_ad_accounts`, `get_campaigns`, `get_campaign`, `get_campaign_groups`, `get_creatives`, `get_ad_account_users`, `get_ad_analytics`
- [x] `approximateMemberReach` returns real values (13005 unique members on 21557 impressions)
- [x] Derived `frequency` correct (6545 / 4141 = 1.5805)
- [x] `audiencePenetration` returns real values
- [x] Lead Gen fields served under the existing scopes, confirming no new scope is required
- [x] 92-day boundary confirmed: reach present at exactly 92 days, silently absent at 574
- [x] Field names validated by control probe — an unknown field returns 400, so a 200 confirms the name exists in the schema
- [x] 63 unit tests pass
- [x] `ruff check` and `ruff format --check` clean

Not yet observed: a non-zero lead value. No accessible test account has ever run a Lead Gen campaign, so those fields correctly return `0`. The plumbing and permissions are verified; the values await an account with form delivery.

## Notes for consumers

LinkedIn omits metrics it has no data for rather than returning `0` — `qualifiedLeads`, `costPerQualifiedLead` and the reach fields are all routinely missing from rows. Every analytics key must be treated as optional. `frequency` is `null` when reach is absent for that row.

Reach cannot be summed across date windows, since it counts unique members and would double-count anyone reached in more than one window.


Shubhank-Jonnada:
All read-only actions were re-tested against a live ad account after the API version bump to 202607, to confirm the version change did not regress anything outside `get_ad_analytics`: `get_ad_accounts`, `get_campaigns`, `get_campaign`, `get_campaign_groups`, `get_creatives`, `get_ad_account_users` and `get_ad_analytics`. All pass.

Unit tests: 63 pass. Integration tests: 14 pass. `ruff check` and `ruff format --check` clean.

The destructive campaign lifecycle test was not run, so the write actions are covered by unit tests only. They are unchanged by this PR apart from the shared version header.

NinosMan:
## Integration PR review

Applied sub-skills: `building-integration`, `writing-unit-tests`, `writing-integration-tests`

### 🚫 Blockers

**`linkedin-ads/linkedin_ads.py:28` — analytics rows come back with no campaign and no date, so none of the new metrics can be attributed.**

`fields` never requests `pivotValues` or `dateRange`. Rest.li field projection is exact: whatever isn't named is dropped. With `pivot=CAMPAIGN` and `timeGranularity=DAILY`, that produces N anonymous rows. Verified live against account `514915345`:

```
rows: 21
keys on row 0: ['appointmentsScheduled', 'clicks', 'costInLocalCurrency',
 'externalWebsiteConversions', 'impressions', 'oneClickLeadFormOpens',
 'oneClickLeads', 'viralOneClickLeadFormOpens', 'viralOneClickLeads']
```

21 rows, no campaign URN, no day. A consumer cannot tell which campaign booked the appointment or which day the reach applies to — which is the entire point of the metrics this PR adds. The docs are explicit: *"LinkedIn requires you to mention specific metrics in a request, using the `fields` query parameter"*, and their own sample names `pivotValues` and `dateRange` alongside the metrics.

The omission predates this PR, but it belongs here: the PR rewrites this exact expression, and the value of the added metrics depends on it.

Fix — append both to `ANALYTICS_FIELDS`. Verified live: 200, 15 fields (under the 20 cap even with reach on), attribution restored:

```
{'dateRange': {'start': {'month': 7, 'day': 1, 'year': 2026}, ...},
 'impressions': 1786, 'pivotValues': ['urn:li:sponsoredCampaign:859109896']}
```

Worth extending `test_returns_analytics` to assert `pivotValues` on returned rows so this can't regress.

### ⚠️ Should-fix

- **`linkedin_ads.py:611`** `include_leadgen_metrics` defaults to `true`, adding 7 metrics to every existing caller's request. LinkedIn throttles on **45M metric values per 5-minute window** (metrics × rows), so this roughly triples existing consumers' data cost against that budget without them opting in. Consider defaulting to `false`, or note the trade-off in the README.
- **`linkedin_ads.py:353`** (pre-existing, destructive action — not run) `create_campaign` never sends `targetingCriteria`, which the campaign schema marks **Required: "True, unless targeting provided"**. Real creates would likely fail.
- **`config.json:159`** (pre-existing) `cost_type` is modelled as optional — the schema table marks `costType` **Required: True**. Conversely `objectiveType` is documented **Required: False** but `config.json:202` requires it.

### 💡 Suggestions

- **`config.json:132`** The docs' field table lists plural objective symbols (`JOB_APPLICANTS`, `WEBSITE_CONVERSIONS`, `WEBSITE_VISITS`, `VIDEO_VIEWS`) while every worked example uses the singulars the config has. The docs contradict themselves; the config's inline note ("singular enum symbols as required by the API") is probably right, but `WEBSITE_TRAFFIC` only applies to `DYNAMIC` campaigns, which the `type` enum excludes — it's unreachable.
- `validate_integration.py` warns `r_ads_reporting` looks unused. It's a false positive (adAnalytics needs it); no action, just noise to expect in CI.

### Validation

**Tooling run** (all green): `validate_integration.py` (1 warning, above) · `check_code.py` — ruff lint/format, bandit, pip-audit, config-code sync, fetch patterns · `check_readme.py` · `check_version_bump.py` (2.0.0 → 2.1.0, minor) · `run_tests.py` — 63/63 unit, 95% coverage.

NinosMan (approved):

PR #429, opened 2026-07-30
fix(github): handle directory paths in get_file_content

Shubhank-Jonnada:
## Problem

An agent using the GitHub integration had no working way to browse a repository's tree. GitHub's contents endpoint (`GET /repos/{owner}/{repo}/contents/{path}`) returns a single object for a file but an **array of entries** for a directory. `get_file_content` only handled the object shape and called `.get()` on the response unconditionally, so any directory path failed with:

```
'list' object has no attribute 'get'
```

Since `get_file_content` is the only action that touches repository contents, that error left directory listing entirely unavailable — the agent could fetch a file only if it already knew the exact path.

## Fix

`get_file_content` now detects the array response and returns a directory listing instead of crashing. The output shape is stable across both cases, so a workflow can branch on `type` rather than guess:

| Field | File response | Directory response |
|---|---|---|
| `type` | `"file"` | `"dir"` |
| `content` | decoded file content | `""` |
| `sha` | file SHA | `""` |
| `size` | bytes | `0` |
| `name` / `path` | file name and path | directory name and path |
| `entries` | `[]` | one object per child |

Each entry carries `name`, `path`, `type` (`file` or `dir`), `sha`, `size`, and `download_url`, so an agent can walk deeper by feeding a child's `path` back into the same action.

The action description in `config.json` now states that a directory path returns a listing, which is what stops the model from treating a folder request as a broken file fetch.

## Changes

- `github/github.py` — `GitHubAPI.get_file_content` branches on the response type and normalises both shapes; handler passes `type` and `entries` through
- `github/config.json` — output schema gains `type` and `entries` (with a typed item schema); action and `path` descriptions document directory behaviour; version `2.5.0` → `2.6.0`
- `github/tests/test_github_misc_unit.py` — unit test covering the directory response, including `download_url: null` normalising to `""`
- `github/tests/test_github_integration.py` — live tests for both shapes against public repos (`octocat/Hello-World` for a file, `github/gitignore` at `Global` for a directory)
- `github/README.md` — `get_file_content` docs rewritten with the full output contract and a note on walking a tree; usage example lists a directory before fetching a file; version history entry added

No new environment variables. The live tests reuse the existing `GITHUB_ACCESS_TOKEN` from the root `.env.example` and hit public repos only, so no extra repo access is needed.

## Validation

- [x] `pytest github/tests` — 104 passed
- [x] `ruff check github/` — clean
- [x] `ruff format --check github/` — clean
- [x] `config.json` parses; `get_file_content` output schema matches the handler's returned keys
- [ ] Live integration tests — collected (7 tests) but skipped locally, no `GITHUB_ACCESS_TOKEN` set in this environment

## Deploying

The agent runs the deployed bundle, so this needs a repackage and upload as 2.6.0 before the directory listing works in production.

### DIFF
--- a/github/README.md
+++ b/github/README.md
@@ -14,7 +14,7 @@ This integration provides complete access to GitHub's REST API (v2022-11-28), al
 - **Commit Operations**: Access commit history and details
 - **Branch Management**: Create, delete, and manage branches
 - **Webhook Integration**: Set up automated workflows with webhooks
-- **File Operations**: Read, create, update, and delete files directly in repositories
+- **File Operations**: Read, create, update, and delete files directly in repositories, and browse directory listings
 - **Gist Management**: Create and manage code snippets
 - **Rate Limiting**: Built-in rate limit monitoring and handling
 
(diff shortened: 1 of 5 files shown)


TheRealAgentK (approved):
LGTM

PR #433, opened 2026-08-03
feat(guidecx): add GUIDEcx integration with 27 actions

Shubhank-Jonnada:
## Author commitment

Before opening an integration PR, I confirm that:

- I understand the third-party API areas I implemented.
- I can explain every action, endpoint, permission, dependency, and important helper.
- I used the documented Autohive SDK process and repository conventions.
- I reviewed the generated or written code myself.
- I removed dead code, generic boilerplate, hallucinated behavior, and unnecessary abstractions.
- I tested the integration with meaningful mocked tests.
- I included real API integration tests where safe and practical, or documented why not.
- I ran local validation and fixed issues before requesting review.
- I documented the integration clearly for users and future maintainers.

If I cannot honestly confirm these points, the PR is not ready.

### Evidence against the charter

Mapped to the [Integration Engineering Charter](https://github.com/Autohive-AI/autohive-integrations/blob/master/integration-engineering-charter.md) so a reviewer can check the claims rather than take them on trust.

**1. We understand the API before we wrap it.** The v3 reference page requires auth, so the implementation is grounded in the public OpenAPI spec at `api.guidecx.com/api/v3/swagger.json`. No endpoint, parameter or response field here was inferred from an example. Where the spec was silent or self-contradictory it was resolved empirically against a live workspace, and the findings are recorded in the integration README: the webhook event type set, the task status labels, the real page-size cap of 500, and the additive tag semantics.

**2. We build for user workflows, not endpoint coverage.** 27 of the API's 42 endpoints are exposed. Templates, CSAT, metadata, messages search, custom fields and billing profiles are deliberately left out for now. `POST /members/embed-token` is excluded on purpose: it mints a credential for embedding GUIDEcx UI, which is not a workflow step.

**3. We own AI-generated code.** This was built with AI assistance. Every provider behaviour claim in the code comments and README was verified against the live API rather than asserted, and three defects found that way are listed below.

**4. We write maintainable Python.** One module, flat handlers, four shared helpers (`prune`, `prune_body`, `paged_result`, `gcx_fetch`). No inheritance beyond the SDK's `ActionHandler`, no dependencies beyond the SDK itself.

**5. We treat auth and permissions as security decisions.** Single credential field, `format: password`, marked required so the SDK rejects a call before any request is made. No secrets in the repo; `.env.example` carries names only. No scopes to over-request, since the token carries its owner's permissions.

**6. We verify behavior, not just syntax.** 125 mocked unit tests plus 41 live end-to-end tests against the real API. All 27 actions have both. Nothing is skipped.

**7. We document what future maintainers need to know.** The integration README covers auth, both token types and how to tell them apart, every action, the status labels, the webhook event types, ten documented provider quirks, and troubleshooting. It describes the integration as shipped.

**8. We self-review before asking others to review.** Local validation is `hiveup validate --changed --base-ref origin/master`, 13/13 checks. The diff contains no stray files: the 19 unrelated `.zip` files at the repo root and an unrelated `fix(github)` commit on the previous branch are both excluded.

### Defects found by testing against the real API

Three bugs that mocked tests alone would not have surfaced:

1. **Optional dates return `null` rather than being omitted.** Output schemas declared plain strings, so `list_projects` failed the SDK's output validation on any real project. It passed earlier only because the workspace had no projects yet.
2. **The SDK rejects `bool` query parameters** (`value should be str, int or float`), so `include_disabled` is now serialised as `"true"` / `"false"`.
3. **The page-size cap is 500, not 100.** `limit=1000` returns `metadata.limit=500`. The original constant was a guess and was silently capping callers at a fifth of what the API allows.

Two further fixes came from review on this PR: preserving empty arrays in request bodies so tags can be cleared, and an explicit fall-through in a test helper.

### Not covered, stated plainly

- `add_task_note` with `internal_only: false`, and customer-visible task `visibility`. Both create customer-facing content that v3 cannot delete.
- Some filters are verified at unit level (correct query parameter emitted) but not against the API: `customer_id`, `project_manager_id`, `tag` and the date ranges on projects; `assignee_id`, `milestone_id` and `name` on tasks.
- `icon.png` is a generated placeholder and must be replaced with the official GUIDEcx logo before this ships.

---

## Summary

GUIDEcx is a customer onboarding and implementation platform where work is organised as projects containing phases, milestones and tasks, each belonging to a customer. Teams running structured onboarding need that progress visible in their other systems, and need to push updates back without leaving the workflow.

This integration covers onboarding reporting, project structure creation, task management, dependencies, the project team, time tracking and webhook subscriptions. Webhooks in particular mean workflows can react to GUIDEcx events rather than polling for them.

Built against the v3 Open API. The reference page at `api.guidecx.com/api/v3/docs` requires auth, but the full OpenAPI spec is public at `api.guidecx.com/api/v3/swagger.json`, so every endpoint, parameter and response shape here came from the spec rather than guesswork, and each was then verified against a live workspace.

**Every one of the 27 actions has been tested locally against the real GUIDEcx API**, not just against mocks. All 27 carry both unit and live end-to-end coverage, with no action untested and no test skipped. Full results are in a comment below.

### Projects

| Action | Description |
|---|---|
| `list_projects` | Search projects by status category, customer, project manager, tag or update date |
| `get_project` | Retrieve one project by ID |
| `create_project` | Create a project, optionally with an inline customer and a project manager |
| `update_project` | Update name, dates, status, tags or contract value |
| `get_customer` | Retrieve a customer by ID |

### Structure

| Action | Description |
|---|---|
| `list_phases` | List a project's phases in display order |
| `list_milestones` | List a project's milestones in display order |
| `list_tasks` | Search tasks by project, milestone, assignee, status category or name |
| `create_phase` | Create a phase in a project |
| `create_milestone` | Create a milestone inside a phase |
| `create_task` | Create a task under a milestone |
| `update_task` | Update a task's status, due date, assignee or priority |
| `add_task_note` | Post a note to a task's message channel |

### Dependencies

| Action | Description |
|---|---|
| `list_dependencies` | List task dependencies by project or by either end |
| `add_dependency` | Make one task depend on another |
| `remove_dependency` | Remove a dependency between two tasks |

### Team

| Action | Description |
|---|---|
| `list_project_members` | List the members assigned to a project, with project role |
| `remove_project_member` | Unassign a member from a project's team |
| `list_members` | Search workspace members by email, role or invitation status |
| `list_roles` | List the roles configured in the workspace |

### Time tracking

| Action | Description |
|---|---|
| `list_time_categories` | List time categories and their billable rates |
| `list_time_records` | Search logged time by project, task, member, category or work date |
| `log_task_time` | Log hours against a task |
| `log_project_time` | Log hours against a project |

### Webhooks

| Action | Description |
|---|---|
| `list_webhooks` | List webhook subscriptions with delivery history |
| `upsert_webhook` | Create or update a webhook subscription |
| `delete_webhook` | Delete a webhook subscription |

## Authentication

- **Type**: Custom
- **Fields**: API Token (password) - a GUIDEcx Open API **user-based token**, generated under Company Settings → Open API by a workspace admin

GUIDEcx also issues legacy **workspace tokens**, which are deprecated and authenticate only against v2. This integration targets v3, so a legacy token returns 401 on every action. The README covers how to tell them apart.

The token inherits its owner's permissions, so it only sees the projects that user can see, and notes and time records are attributed to that user.

## Provider behaviour worth knowing

Documented in the integration README; summarised here because several of these shaped the action design.

- **There is no `GET /projects/{id}`.** The v3 API offers only a search endpoint with a repeatable `id` filter, so `get_project` performs a single-ID search and converts an empty result into an error rather than an empty success.
- **Tasks cannot hang off a project.** The hierarchy is phase → milestone → task, so `create_task` requires `milestone_id`; omitting it fails with `task [0] has an invalid parentId: ""`.
- **Task and project writes go through batch upserts.** Verified against the live API to be genuine partial updates - a priority-only change left `name`, `status`, `responsibility`, `visibility` and `tags` untouched - despite the endpoint being named "upsert". There is a test asserting this.
- **Optional dates return `null` rather than being omitted**, so the output schemas accept null. A schema declaring them as plain strings fails the SDK's output validation as soon as a real record comes back.
- **Webhook event types have no enum in the spec**, and its two examples contradict each other (`task.updated` in the request, `TASK_UPDATED_EVENT` in the response). The accepted set was determined empirically and is now enumerated in `config.json`. Notably there is no completion event - subscribe to `task.updated` and check `statusCategory`.
- **The webhook endpoints use snake_case** for both query parameters and response fields, unlike the rest of the API, and return `event_type` as `null` even on a successful upsert.
- **The page-size cap is 500, not 100.** A larger `limit` is accepted but silently clamped (`limit=1000` returns `metadata.limit=500`), so the schema rejects anything above 500 rather than letting a caller believe they received a full result set.
- **Nothing structural can be deleted.** v3 has no delete endpoint for projects, phases, milestones, tasks, messages or time records. Only dependencies and webhooks are removable, and `remove_project_member` unassigns rather than deletes.
- **Projects created in the classic GUIDEcx UI are invisible to the v3 API**, returning nothing even when queried by exact ID, while remaining visible in the UI to the same admin user. Projects created in GUIDE 2.0 appear immediately. Recorded as a troubleshooting entry since it presents as a missing-data bug.

## Files added

- `guidecx/config.json` - schema, auth and all 27 action definitions
- `guidecx/guidecx.py` - action handlers
- `guidecx/__init__.py` - re-export
- `guidecx/requirements.txt` - `autohive-integrations-sdk~=2.0.1`
- `guidecx/README.md` - auth setup, actions, status labels, event types, provider behaviour, troubleshooting
- `guidecx/icon.png` - 512×512 icon
- `guidecx/tests/test_guidecx_unit.py` - 115 mocked unit tests
- `guidecx/tests/test_guidecx_integration.py` - 40 live end-to-end tests
- `guidecx/tests/{__init__,conftest}.py` - test scaffolding

Also updated the root `README.md` and `.env.example`.

## Validation

`hiveup validate --changed --base-ref origin/master` - all 13 checks pass:

structure, syntax, imports, json, lint, format, security, audit, sync, fetch, tests, readme, version

## Testing

| Suite | Result |
|---|---|
| Unit (`pytest -m unit`) | 115 passed |
| E2E read-only (`-m "integration and not destructive"`) | 29 passed, 0 skipped |
| E2E destructive (`-m "integration and destructive"`) | 11 passed |

All 27 actions have both unit and live end-to-end coverage; no action is untested and nothing skips.

The destructive suite is opt-in, excluded from CI by both the marker filter and the `python_files` pattern. It cleans up where the API permits: task tests restore the original status and priority, and the webhook lifecycle test deletes what it created, pointing at `example.com` (reserved by RFC 2606) so no stray delivery can reach a real host. Where v3 has no delete endpoint it minimises impact instead - notes are posted `internal_only`, and time entries are 0.01 hours.

### Reviewer notes

- `icon.png` is a generated placeholder and should be replaced with the official GUIDEcx logo before this ships.
- Two paths are deliberately untested: `add_task_note` with `internal_only: false`, and customer-visible task `visibility`. Both would create customer-facing content that cannot be deleted.
- Endpoint coverage is intentionally partial. Templates, CSAT, metadata, messages search, custom fields and billing profiles are not exposed yet. `POST /members/embed-token` is deliberately excluded: it mints a credential for embedding GUIDEcx UI, which is not a workflow step.

---

## Review standard

Reviewers are expected to hold integration PRs to this charter.

A PR may be sent back even if CI passes when the author has not demonstrated API understanding, intentional scope, safe permissions, meaningful tests, maintainable Python, or adequate self-review.

The goal is not to slow down integration development. The goal is to decrease time to develop and maintenance by avoiding shipping code we do not understand, cannot maintain long term, or cannot safely stand behind.

If you are unsure about anything on the above to help you move forward in the integration, reach out to a peer to get clarity


Shubhank-Jonnada:
## Local test results

| Suite | Result |
|---|---|
| Unit | **115 passed** |
| E2E read-only | **29 passed, 0 skipped** |
| E2E destructive | **11 passed** |

`hiveup validate --changed --base-ref origin/master` — **13/13 passed**: structure, syntax, imports, json, lint, format, security, audit, sync, fetch, tests, readme, version.

All 27 actions were tested against the real GUIDEcx v3 API. Every action has both unit and live end-to-end coverage; none untested, none skipped.

Shubhank-Jonnada:
@codex review

guidecx/tests/test_guidecx_integration.py
```diff
@@ -0,0 +1,862 @@
...
+        project_id = await resolve_project_id(live_context)
+
+        result = await guidecx.execute_action(
+            "list_tasks", {"project_id": project_id, "type_filter": "TASK"}, live_context
+        )
+
+        assert result.type == ResultType.ACTION, result.result
+
+
+async def resolve_customer_id(live_context):
```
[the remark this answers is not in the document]
Shubhank-Jonnada: Added an explicit raise after pytest.skip so there is no implicit fall through return. Used a raise rather than the suggested `return ""` because an empty string is a plausible looking customer id that would reach get_customer and surface as a 404 instead of a clean skip - https://github.com/Autohive-AI/autohive-integrations/commit/2479c1a780072e558416b48f0ebe0249073316b5

guidecx/guidecx.py
```diff
@@ -0,0 +1,880 @@
...
+            project_id = inputs["project_id"]
+
+            project = prune(
+                {
+                    "name": inputs.get("name"),
+                    "startDate": inputs.get("start_date"),
+                    "endDate": inputs.get("end_date"),
+                    "status": inputs.get("status"),
+                    "statusExplanation": inputs.get("status_explanation"),
+                    "tags": inputs.get("tags"),
```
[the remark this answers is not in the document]
Shubhank-Jonnada: Split prune() so request bodies keep empty arrays: prune() still drops them for query params, new prune_body() only drops None and empty string. Verified on the live API that `tags: []` clears tags and an omitted key leaves them untouched. Also corrected the tags description, tags are additive not a replacement set and only an empty array removes anything - https://github.com/Autohive-AI/autohive-integrations/commit/c8b4d624746ac20b42d8fc0bd4a4bce44780d826

Shubhank-Jonnada:
@codex review

TheRealAgentK:
`guidecx/guidecx.py` passes list-valued query parameters directly to `context.fetch(params=...)`, but `autohive-integrations-sdk` 2.0.1 JSON-serializes lists before URL encoding. GUIDEcx declares these filters with `collectionFormat: multi`, so it expects repeated parameters such as `statusCategory=IN_PROGRESS&statusCategory=LATE`; the SDK instead produces one JSON-array value.

This affects `get_project` and the list-valued project, task, member, time-record, and dependency filters. The live tests do not exercise the production behavior because their `real_fetch` fixture passes `params` directly to `aiohttp`, which serializes lists differently.

Encode query parameters with `urlencode(params, doseq=True)` before calling `context.fetch`, and add a test that verifies the final query string uses repeated parameters.

TheRealAgentK:
Several actions insert caller-provided IDs directly into URL paths. For example:

```python
await gcx_fetch(context, "DELETE", f"/webhooks/{webhook_id}")
```

The code assumes `webhook_id` is a UUID, but `config.json` only requires a string. A caller can therefore provide path syntax instead of an ID:

```text
../projects/PROJECT_ID/members/MEMBER_ID
```

That constructs:

```text
https://api.guidecx.com/api/v3/webhooks/../projects/PROJECT_ID/members/MEMBER_ID
```

The HTTP URL library normalizes `..` before sending the request, resulting in:

```text
DELETE https://api.guidecx.com/api/v3/projects/PROJECT_ID/members/MEMBER_ID
```

This allows `delete_webhook` to call the API route for removing a project member, breaking the action boundary. The same direct interpolation pattern is used for task messages, time records, project members, phases, milestones, and tasks.

Validate every path-bound ID as a canonical GUIDEcx UUID before constructing the URL. Add UUID regex patterns to the relevant `config.json` inputs and runtime validation in a shared helper. Adding only `"format": "uuid"` is insufficient because the current SDK's JSON Schema validator does not enforce formats.

guidecx/guidecx.py
```diff
@@ -0,0 +1,899 @@
...
+            if not project:
+                return ActionError(
+                    message=(
+                        "No update fields provided. Set at least one of name, start_date, end_date, "
+                        "status, status_explanation, tags or cash_value."
+                    )
+                )
+            project["id"] = project_id
+
+            body = await gcx_fetch(context, "PATCH", "/projects", json_body={"projects": [project]})
```
TheRealAgentK, line 809: `update_project` sends the caller-provided ID directly to GUIDEcx's batch-upsert endpoint without first verifying that the project exists.

GUIDEcx documents `PATCH /api/v3/projects` in its [public v3 OpenAPI specification](https://api.guidecx.com/api/v3/swagger.json) as:

> Creates or updates multiple projects in a single request. If a project with the given ID exists, it will be updated. Otherwise, a new project will be created.

This means a typo or stale `project_id` passed to an action described as update-only can create a new project instead. GUIDEcx v3 has no project deletion endpoint, so that accidental creation cannot be cleaned up through this integration.

Search for the exact project ID before this PATCH and return `ActionError` unless it exists and exactly matches. This existence check should use the corrected repeated-query serialization discussed in the other comment.

guidecx/guidecx.py
```diff
@@ -0,0 +1,899 @@
...
+    """Create a phase in a project."""
+
+    async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
+        try:
+            project_id = inputs["project_id"]
+            phase = prune_body(
+                {
+                    "name": inputs["name"],
+                    "templateId": inputs.get("template_id"),
+                    "formattedDescription": inputs.get("description"),
```
TheRealAgentK: `create_phase` sends `formattedDescription`, but GUIDEcx's [public v3 OpenAPI specification](https://api.guidecx.com/api/v3/swagger.json) defines `phaseUpsertPhaseInput` with only `id`, `name`, `templateId`, and `position`. Unlike the milestone and task upsert schemas, it has no phase-description field.

The unit test only verifies that this payload is constructed, and the live `create_phase` test does not supply a description, so neither establishes that GUIDEcx accepts or persists this field. Remove `description` from the action schema and request until a live create-and-read-back test proves that GUIDEcx supports it.

guidecx/guidecx.py
```diff
@@ -0,0 +1,899 @@
...
+            )
+
+            # A customer is either referenced by ID or created inline from a
+            # name and domain. There is no dedicated create-customer endpoint,
+            # so the inline form is the only way to make one via the API.
+            customer_id = inputs.get("customer_id")
+            if customer_id:
+                project["customer"] = {"id": customer_id}
+            elif inputs.get("customer_name"):
+                project["customer"] = prune({"name": inputs["customer_name"], "domain": inputs.get("customer_domain")})
```
TheRealAgentK: GUIDEcx's [public v3 OpenAPI specification](https://api.guidecx.com/api/v3/swagger.json) states for `projectUpsertProjectInputCustomer`: "If an id is not provided, the name and domain must be provided."

The integration currently allows `customer_name` without `customer_domain`, passes local schema validation, and sends a name-only customer object that violates that contract. `customer_domain` without `customer_name` is also silently ignored.

Require `customer_name` and `customer_domain` together whenever `customer_id` is absent, using both input-schema validation and a handler check before the request.

guidecx/guidecx.py
Shubhank-Jonnada: update_project now looks the exact ID up via GET /projects before the upsert and returns ActionError if it does not exist, verified live that an unknown ID gets refused with no PATCH sent, and get_project reuses the same lookup so it no longer trusts data[0] - https://github.com/Autohive-AI/autohive-integrations/commit/67f5f092e33f4cf13f904dff0f4a534bb6849598

guidecx/guidecx.py
Shubhank-Jonnada: Removed description from the create_phase schema, handler and README, confirmed live that a phase created with formattedDescription comes back without it in both the create response and the read-back from GET /phases, while a milestone control in the same run does echo it and a bogus field also returns 200 - https://github.com/Autohive-AI/autohive-integrations/commit/cfbe2e612a4a30be388c00bb99216b1222200aa4

guidecx/guidecx.py
Shubhank-Jonnada: customer_name and customer_domain are now required together whenever customer_id is absent, enforced in the input schema with Draft 7 if/then dependencies and repeated in the handler, confirmed live that a name-only customer is rejected by the API with 400 invalid customer so that path could never have worked - https://github.com/Autohive-AI/autohive-integrations/commit/bdf0db7c8f47e75f26f1833c86d347e5d1d8314b

Shubhank-Jonnada:
Fully tested locally after the three review fixes above.

| Suite | Result |
|---|---|
| Unit | **133 passed** |
| E2E read-only | **29 passed, 0 skipped** |
| E2E destructive | **12 passed** |
| `hiveup validate` | **passed** |
| `ruff check` / `ruff format` | **clean** |

Each fix was also checked directly against the live v3 API before replying:

| Fix | Live check |
|---|---|
| Project existence check | Unknown ID refused with no PATCH sent, real ID still updates |
| Phase description removed | Phase created with `formattedDescription` returns 200 but comes back without it, in the create response and on read-back from `GET /phases`; a milestone control in the same run does echo it |
| Customer name and domain pairing | Name-only customer rejected by the API with `400 invalid customer`, so that path could never have worked |

Shubhank-Jonnada:
Both confirmed and fixed, one commit each.

**Repeated query parameters** ([507d93f](https://github.com/Autohive-AI/autohive-integrations/commit/507d93f91e4a4b916bc43885734d975dc14cf2fa))

Verified against the live API, and it is worse than "wrong serialization": GUIDEcx returns **HTTP 400** for the JSON-array form.

```
id=%5B%2221b3657e-...%22%5D  ->  HTTP 400
id=21b3657e-...             ->  HTTP 200, 1 project
```

So `get_project` has never worked in production, and the project existence check I added earlier in this PR would have made `update_project` fail outright rather than protect anything. `gcx_fetch` now builds the query string with `build_query` and appends it to the URL, so lists become repeated parameters and booleans are lowercased instead of `str(True)`.

Your point about the fixture was the important one. It forwarded `params` straight to aiohttp, which serializes lists as repeated pairs, so the e2e suite was exercising a serialization production never used. The fixture now asserts `params is None`, which makes that divergence impossible to reintroduce.

**UUID validation on path-bound IDs** ([1d3cc22](https://github.com/Autohive-AI/autohive-integrations/commit/1d3cc22b7c790a3a0041b0046637b9d0e64e278f))

Reproduced the rewrite with yarl against the real base URL:

```
/api/v3/webhooks/../projects/<pid>/members/<mid>
  -> /api/v3/projects/<pid>/members/<mid>
```

`path_id()` validates every ID that reaches a path across all eleven actions, with a matching `pattern` in `config.json`. Agreed on `format: uuid` being insufficient, so the handler check is the one that has to hold and there is a test asserting the handler refuses traversal even when called directly, bypassing schema validation. A test also pins the yarl rewrite itself so the risk does not become theoretical.

**Results**

| Suite | Result |
|---|---|
| Unit | **163 passed** |
| E2E read-only | **29 passed** |
| E2E destructive | **12 passed** |
| `hiveup validate` | **passed** |
| `ruff check` / `ruff format` | **clean** |

TheRealAgentK (approved):
Thx for the changes, @Shubhank-Jonnada - LGTM now

PR #438, opened 2026-08-04
fix(netlify): normalize deploy file paths to prevent 404 on upload

Shubhank-Jonnada:
## Summary

Every `create_deploy` call whose `files` keys lacked a leading slash failed with `HTTP 404: Not Found`. The upload URL was built by concatenating the file path straight onto the endpoint, so `index.html` produced `.../deploys/<id>/filesindex.html` instead of `.../deploys/<id>/files/index.html` — a route that doesn't exist. Nothing in the input schema said the leading slash was required, so writing `index.html` was the natural way to call the action, and it never worked.

The failure was easy to misdiagnose: reads (`list_sites`, `get_site`) succeed against the same site and token, so it looks like a permissions or scope problem rather than a string bug.

## The fix

Paths are normalized to exactly one leading slash before being used in two places:

- the digest map POSTed to `/sites/{site_id}/deploys`
- the `PUT` upload URL for each required file

Both had to change together. Netlify keys the `required` hash list off whatever paths you declare, so normalizing only the URL would leave files silently unuploaded — a deploy that reports success with no content. `index.html` and `/index.html` now produce byte-identical requests.

## Error reporting

`create_deploy` makes three requests and the handler collapsed any failure to `str(e)`, so a 404 gave no indication of which one broke. The action now tracks the in-flight step and includes it:

```
HTTP 404: Not Found (while uploading /index.html to deploy 6899a1f2)
```

## Tests

**Unit** — four cases added to `netlify/tests/test_netlify_unit.py`:

| Test | Covers |
|------|--------|
| `test_create_deploy_path_without_leading_slash_is_normalized_in_digests` | Unslashed path is declared to Netlify as `/index.html` |
| `test_create_deploy_path_without_leading_slash_uploads_to_valid_url` | Regression: asserts `filesindex.html` never appears in the upload URL |
| `test_create_deploy_normalizes_leading_slash_consistently` | Slashed and unslashed inputs produce identical request URLs |
| `test_create_deploy_error_message_names_the_failing_step` | Error message identifies the failing request |

The existing suite only ever passed `"/index.html"`, which is why this never surfaced in CI.

**Integration** — three destructive cases added in `TestDeployFilePathNormalization`:

| Test | Covers |
|------|--------|
| `test_deploy_succeeds_with_path_missing_leading_slash` | The exact call that used to 404 now succeeds against the real API |
| `test_deployed_file_is_actually_served` | Fetches the live `deploy_url` and asserts the uploaded content is in the response |
| `test_slashed_and_unslashed_paths_both_deploy` | Slashed, unslashed, and nested paths in one multi-file deploy |

`test_deployed_file_is_actually_served` is the one that matters. A digest/upload-path mismatch yields a deploy that reports success while serving nothing, which an ID-only assertion cannot catch — so this asserts on the served page, retrying while Netlify propagates the deploy.

Also added a `result.type` assertion to the pre-existing `TestDeployLifecycle`, which previously raised `AttributeError` on failure instead of surfacing the API's error message.

## Docs

- `config.json` and `README.md`: `files` maps site-root-relative paths, leading slash optional
- `README.md`: corrected the "paths should start with `/`" note, which is now wrong
- `README.md`: replaced the stale testing instructions (they pointed at a `tests/test_netlify.py` that doesn't exist and said to hardcode tokens into test files) with the real commands — read-only first, destructive second with a warning

No `.env.example` change needed: `NETLIFY_ACCESS_TOKEN` is already documented, and the new tests introduce no new env vars.

## Validation

- [x] `pytest netlify/tests/` — 46 passed
- [x] Integration tests collect under `-m integration` (13) and are excluded from the default run
- [x] `ruff check` — clean
- [x] `ruff format` — applied
- [x] `config.json` — parses

The integration tests were not executed — they need a real `NETLIFY_ACCESS_TOKEN`, and the new ones create and delete real sites. To verify locally:

```bash
# read-only
pytest netlify/tests/test_netlify_integration.py -m "integration and not destructive"

# creates and deletes real sites on the connected account
pytest netlify/tests/test_netlify_integration.py -m "integration and destructive"
```

### DIFF
--- a/netlify/netlify.py
+++ b/netlify/netlify.py
@@ -1,5 +1,6 @@
 from autohive_integrations_sdk import Integration, ExecutionContext, ActionHandler, ActionResult, ActionError
 from typing import Dict, Any
+from urllib.parse import quote
 import hashlib
 
 # Create the integration
@@ -7,6 +8,73 @@ netlify = Integration.load()
 
 # Base URL for Netlify API
 NETLIFY_API_BASE_URL = "https://api.netlify.com/api/v1"
+
+# Netlify's API guide is explicit: "Be sure to escape the file_path parameter,
+# and ensure file paths don't include # or ? characters." A path is used two
+# different ways, and conflating them is what makes this subtle:
+#
+#   * as a *logical* key in the deploy's file digest ("/index.html"), unescaped
+#   * as part of the *upload URL* (PUT /deploys/{id}/files/index.html), escaped
+#
+# Anything the URL parser treats specially therefore has to be rejected or
+# encoded, or the two disagree and the upload silently targets the wrong path.
+FORBIDDEN_PATH_CHARACTERS = {
+    "#": "fragment delimiter",
+    "?": "query delimiter",
+    "\\": "backslash",
+}
+
+
+def normalize_deploy_path(path: Any) -> str:
+    """Validate a caller-supplied deploy path and return its logical form.
+
+    The result always has exactly one leading slash and no dot segments, so it
+    is safe to use as a digest key and, once escaped, as an upload URL.
+
+    Raises ValueError for anything that could not round-trip: an empty path, a
+    character the URL parser would treat as a delimiter, or a component that
+    URL normalization would rewrite. Dot segments matter most here: an upload
+    URL containing "/../" is normalized by the HTTP client before it is sent,
+    so "assets/../index.html" would be declared under that literal digest key
+    but uploaded to "/files/index.html", and "../secret.txt" would resolve
+    outside the /files endpoint entirely.
+    """
+    if not isinstance(path, str) or not path.strip():
+        raise ValueError("Deploy file paths cannot be empty.")
+
+    for character, description in FORBIDDEN_PATH_CHARACTERS.items():
+        if character in path:
+            raise ValueError(
+                f"Deploy file path '{path}' contains '{character}' ({description}), "
+                "which Netlify does not allow in deploy paths."
+            )
+
+    components = path.lstrip("/").split("/")
+    if not any(components):
+        raise ValueError("Deploy file paths cannot be empty.")
+
+    for component in components:
+        if component == "":
+            raise ValueError(f"Deploy file path '{path}' has an empty path segment (consecutive slashes).")
+        if component in (".", ".."):
+            raise ValueError(
+                f"Deploy file path '{path}' contains a '{component}' segment. "
+                "Relative segments are rewritten by URL normalization, so the uploaded "
+                "file would not land at the declared path."
+            )
+
+    return "/" + "/".join(components)
+
+
+def encode_deploy_path(logical_path: str) -> str:
+    """Percent-encode a normalized path for use in the upload URL.
+
+    Each component is escaped separately so that the "/" separators survive,
+    while spaces, non-ASCII characters and literal "%" are encoded. The digest
+    key keeps the unescaped logical form; only the URL is escaped.
+    """
+    return "/" + "/".join(quote(component, safe="") for component in logical_path.lstrip("/").split("/"))
+
 
 # Note: Authentication is handled automatically by the platform OAuth integration.
 # The context.fetch method automatically includes the OAuth token in requests.
@@ -143,12 +211,15 @@ class CreateDeployAction(ActionHandler):
             hash_to_content = {}
             hash_to_path = {}
 
+            # Every path is validated before the deploy is created, so an invalid
+            # path fails without leaving a half-built deploy behind.
             for path, content in files.items():
-                # Netlify keys the "required" list off the paths declared here, and its
-                # upload route is /files/<path>. Normalize to a single leading slash so
-                # "index.html" and "/index.html" are treated identically.
-                normalized_path = "/" + path.lstrip("/")
+                # Netlify keys the "required" list off the paths declared here, and
+                # its upload route is /files/<path>. Both sides use this logical
+                # form, so "index.html" and "/index.html" are treated identically.
+                normalized_path = normalize_deploy_path(path)
                 sha1 = hashlib.sha1(content.encode(), usedforsecurity=False).hexdigest()  # nosec B324
+
                 files_dict[normalized_path] = sha1
                 if sha1 not in hash_to_content:
                     hash_to_content[sha1] = content
@@ -172,9 +243,10 @@ class CreateDeployAction(ActionHandler):
                     file_content = hash_to_content[sha1_hash]
                     file_path = hash_to_path[sha1_hash]
 
+                    # The digest key stays unescaped; only the URL is escaped.
                     step = f"uploading {file_path} to deploy {deploy_id}"
                     await context.fetch(
-                        f"{NETLIFY_API_BASE_URL}/deploys/{deploy_id}/files{file_path}",
+                        f"{NETLIFY_API_BASE_URL}/deploys/{deploy_id}/files{encode_deploy_path(file_path)}",
                         method="PUT",
                         headers={"Content-Type": "application/octet-stream"},
                         data=file_content.encode(),
--- a/netlify/netlify.py
+++ b/netlify/netlify.py
@@ -132,6 +132,8 @@ class CreateDeployAction(ActionHandler):
     """Create a new deploy for a site with files."""
 
     async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
+        # Tracks which request is in flight so a failure names the step that broke.
+        step = "preparing file digests"
         try:
             site_id = inputs["site_id"]
             files = inputs["files"]
@@ -142,13 +144,18 @@ class CreateDeployAction(ActionHandler):
             hash_to_path = {}
 
             for path, content in files.items():
+                # Netlify keys the "required" list off the paths declared here, and its
+                # upload route is /files/<path>. Normalize to a single leading slash so
+                # "index.html" and "/index.html" are treated identically.
+                normalized_path = "/" + path.lstrip("/")
                 sha1 = hashlib.sha1(content.encode(), usedforsecurity=False).hexdigest()  # nosec B324
-                files_dict[path] = sha1
+                files_dict[normalized_path] = sha1
                 if sha1 not in hash_to_content:
                     hash_to_content[sha1] = content
-                    hash_to_path[sha1] = path
+                    hash_to_path[sha1] = normalized_path
 
             # Create deploy with file digests
+            step = f"creating deploy for site {site_id}"
             deploy_response = await context.fetch(
                 f"{NETLIFY_API_BASE_URL}/sites/{site_id}/deploys", method="POST", json={"files": files_dict}
             )
@@ -165,6 +172,7 @@ class CreateDeployAction(ActionHandler):
                     file_content = hash_to_content[sha1_hash]
                     file_path = hash_to_path[sha1_hash]
 
+                    step = f"uploading {file_path} to deploy {deploy_id}"
                     await context.fetch(
                         f"{NETLIFY_API_BASE_URL}/deploys/{deploy_id}/files{file_path}",
                         method="PUT",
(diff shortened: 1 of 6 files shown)


netlify/tests/test_netlify_unit.py
```diff
@@ -423,6 +423,82 @@ async def test_create_deploy_multiple_files_hashed_correctly():
...
+        import hashlib
+
+        sha1 = hashlib.sha1(content.encode(), usedforsecurity=False).hexdigest()
+        ctx = make_ctx_multi([{"id": "d1", "required": [sha1]}, None, {"id": "d1", "url": "https://d1.netlify.app"}])
+        return ctx, netlify_integration.execute_action(
+            "create_deploy", {"site_id": "s1", "files": {path: content}}, ctx
+        )
+
+    slashed_ctx, slashed = run("/index.html")
+    await slashed
```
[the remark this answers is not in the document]
Shubhank-Jonnada: Fixed in cffe9be.

Confirmed the finding: the `run()` helper returned an un-awaited coroutine, so `await slashed` was a statement whose result was never consumed.

Rather than assigning throwaway variables, I made the helper `async` so it awaits internally and returns the context it used:

```python
async def deploy(path):
    ...
    result = await netlify_integration.execute_action(...)
    assert result.type == ResultType.ACTION, getattr(result.result, "message", "")
    return ctx

slashed_ctx = await deploy("/index.html")
plain_ctx = await deploy("index.html")
```

That removes the bare-await statements entirely instead of suppressing the symptom, and picks up the suggested success assertion.

One correction on the suggested changeset: `ResultType.JSON` does not exist. The enum members are `ACTION`, `ACTION_ERROR`, `CONNECTED_ACCOUNT`, `ERROR`, `VALIDATION_ERROR`, so the success member is `ResultType.ACTION`. Checking that also surfaced a real bug in my own integration tests, which asserted on a nonexistent `ResultType.ACTION_RESULT` and would have raised `AttributeError` on first run; fixed separately in 1c30434.

netlify/tests/test_netlify_unit.py
```diff
@@ -423,6 +423,82 @@ async def test_create_deploy_multiple_files_hashed_correctly():
...
+    plain_ctx, plain = run("index.html")
+    await plain
```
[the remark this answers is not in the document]
Shubhank-Jonnada: Fixed in cffe9be.

Confirmed the finding: the `run()` helper returned an un-awaited coroutine, so `await slashed` was a statement whose result was never consumed.

Rather than assigning throwaway variables, I made the helper `async` so it awaits internally and returns the context it used:

```python
async def deploy(path):
    ...
    result = await netlify_integration.execute_action(...)
    assert result.type == ResultType.ACTION, getattr(result.result, "message", "")
    return ctx

slashed_ctx = await deploy("/index.html")
plain_ctx = await deploy("index.html")
```

That removes the bare-await statements entirely instead of suppressing the symptom, and picks up the suggested success assertion.

One correction on the suggested changeset: `ResultType.JSON` does not exist. The enum members are `ACTION`, `ACTION_ERROR`, `CONNECTED_ACCOUNT`, `ERROR`, `VALIDATION_ERROR`, so the success member is `ResultType.ACTION`. Checking that also surfaced a real bug in my own integration tests, which asserted on a nonexistent `ResultType.ACTION_RESULT` and would have raised `AttributeError` on first run; fixed separately in 1c30434.

Shubhank-Jonnada:
Testing has been done locally, on the beta, and on the pod ( our plan )

netlify/netlify.py
```diff
@@ -142,13 +144,18 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
             hash_to_path = {}
 
             for path, content in files.items():
+                # Netlify keys the "required" list off the paths declared here, and its
+                # upload route is /files/<path>. Normalize to a single leading slash so
+                # "index.html" and "/index.html" are treated identically.
+                normalized_path = "/" + path.lstrip("/")
```
TheRealAgentK: Adding a leading slash does not make this a safe deploy path. Reserved characters and dot segments still make the digest path differ from the effective PUT URL. For example, `assets/../index.html` is declared under that literal digest key, but the HTTP client normalizes the upload URL to `/files/index.html`; `?` and `#` are parsed as query and fragment delimiters. This change also turns unslashed `../secret.txt` from an invalid `files../secret.txt` route into `/files/../secret.txt`, which normalizes away from the files endpoint.

Netlify's [official API documentation](https://docs.netlify.com/api-and-cli-guides/api-guides/get-started-with-api/#file-digest-method) requires escaping `file_path` and forbids `#` and `?` in file paths.

Validate before creating the deploy: strip leading slashes; reject empty paths, `#`, `?`, backslashes, and empty/`.`/`..` components; retain the validated logical path in the digest; and percent-encode each component separately for the PUT URL while preserving `/` separators.

netlify/netlify.py
```diff
@@ -142,13 +144,18 @@ async def execute(self, inputs: Dict[str, Any], context: ExecutionContext):
...
                 sha1 = hashlib.sha1(content.encode(), usedforsecurity=False).hexdigest()  # nosec B324
-                files_dict[path] = sha1
+                files_dict[normalized_path] = sha1
```
TheRealAgentK, line 235: Normalization introduces a collision: `{"index.html": "first", "/index.html": "second"}` maps both inputs to `/index.html`, and this assignment silently discards the first file. Reversing input order changes the deployed content.

Detect an existing normalized destination before assignment. Coalesce duplicates only when their hashes match; otherwise return `ActionError` identifying the conflicting path.

netlify/tests/test_netlify_integration.py
```diff
@@ -274,3 +296,114 @@ async def test_create_deploy_and_verify(self, live_context):
...
+            )
+
+            assert deploy_result.type == ResultType.ACTION, getattr(deploy_result.result, "message", "")
+            data = deploy_result.result.data
+            assert data["deploy"]["id"]
+            assert data["deploy_url"]
+
+        finally:
+            if site_id:
+                await netlify_integration.execute_action("delete_site", {"site_id": site_id}, live_context)
```
TheRealAgentK: This cleanup result is ignored, as are the cleanup results at the end of the other two new path-normalization tests. `delete_site` catches HTTP failures and returns `ActionError`, so a failed deletion does not raise and the destructive test can pass while leaving a real site behind.

Assert that cleanup returns `ResultType.ACTION` and include the returned error message on failure. Apply the same check to all three new `finally` blocks, preferably through a shared cleanup helper.

netlify/tests/test_netlify_unit.py
```diff
@@ -423,6 +423,83 @@ async def test_create_deploy_multiple_files_hashed_correctly():
...
+        assert result.type == ResultType.ACTION, getattr(result.result, "message", "")
+        return ctx
+
+    slashed_ctx = await deploy("/index.html")
+    plain_ctx = await deploy("index.html")
+
+    def urls(ctx):
+        return [c.args[0] if c.args else c.kwargs.get("url", "") for c in ctx.fetch.call_args_list]
+
+    assert urls(slashed_ctx) == urls(plain_ctx)
```
TheRealAgentK, line 495: These normalization tests cover ordinary leading-slash cases but not boundaries where the digest path and upload URL diverge. Add cases for:

- spaces, Unicode, and literal `%`: logical digest key remains unencoded while the PUT URL is percent-encoded;
- `?`, `#`, empty paths, dot segments, and backslashes: rejected before the deploy POST;
- `index.html` and `/index.html` with conflicting content: rejected rather than silently overwritten.

netlify/netlify.py
Shubhank-Jonnada: Confirmed all four cases with httpx.URL against the real base URL, including that my leading-slash change turned `../secret.txt` into a path resolving outside /files, so normalize_deploy_path now rejects empty paths, #, ?, backslashes, empty segments and ./.. segments before the deploy POST, and encode_deploy_path percent-encodes each component separately for the URL while the digest keeps the unescaped logical path - https://github.com/Autohive-AI/autohive-integrations/commit/e8cd6a4681c86b78c033af14e1cd68b41967bc60

netlify/netlify.py
Shubhank-Jonnada: Collisions after normalization are now detected before assignment, coalesced when the content hashes match and refused with an ActionError naming the conflicting path when they differ, so input ordering can no longer decide which file gets deployed - https://github.com/Autohive-AI/autohive-integrations/commit/6f5d00b28d5700bae209c122fdc920b6ee6bbf64

netlify/tests/test_netlify_integration.py
Shubhank-Jonnada: All five cleanups now go through a shared cleanup_site helper that asserts the result is ACTION and that deleted is true and surfaces the API message on failure, and I also added 429 backoff to the live fetch helper since Netlify throttles site creation to about two per minute and the suite was failing on that rather than on anything real - https://github.com/Autohive-AI/autohive-integrations/commit/54131433b2385ba3c1b1bcba05ecfb66241a309e

netlify/tests/test_netlify_unit.py
Shubhank-Jonnada: Added all three groups, split across the two fix commits they belong to: spaces, Unicode and literal % assert the digest key stays unencoded while the PUT URL is percent-encoded, ?, #, empty, dot-segment and backslash paths assert rejection with no requests sent, and colliding paths with conflicting content assert rejection, plus a test that the escaped URL survives httpx parsing unchanged and live e2e coverage that space-containing paths are actually served - https://github.com/Autohive-AI/autohive-integrations/commit/e8cd6a4681c86b78c033af14e1cd68b41967bc60

Shubhank-Jonnada:
Fully tested locally after the three review fixes above, against the real Netlify API.

| Suite | Result |
|---|---|
| Unit | **92 passed** |
| E2E read-only | **8 passed** |
| E2E destructive | **7 passed** |
| `hiveup validate` | **passed** |
| `ruff check` / `ruff format` | **clean** |

Live evidence for each fix:

| Fix | Live check |
|---|---|
| Path validation and escaping | `my page.html` and `assets/deep dir/style sheet.css` deploy and are served at their logical paths via the percent-encoded URLs; `../secret.txt`, `assets/../index.html`, `a?b.html` and `a#b.html` are all rejected and `list_deploys` confirms no deploy was created |
| Collision detection | Conflicting content under `index.html` and `/index.html` is refused; identical content coalesces to one digest entry |
| Cleanup assertions | Full destructive run leaves no `autohive-*` sites on the account, verified by listing sites afterwards |

One thing worth flagging for anyone running these: Netlify throttles site and deploy creation to roughly two per minute, and the destructive suite creates a site per test. Without backoff the run fails with `429 API Deploy rate limit surpassed` rather than anything real, so the live fetch helper now retries on 429. The destructive run takes about 2m15s as a result.

netlify/tests/test_netlify_unit.py
```diff
@@ -423,6 +430,279 @@ async def test_create_deploy_multiple_files_hashed_correctly():
...
+
+    upload_url = request_urls(ctx)[1]
+    assert upload_url.endswith(expected_url_suffix)
+
+
+@pytest.mark.asyncio
+async def test_create_deploy_preserves_slash_separators_when_escaping():
+    """Separators must survive encoding, or nested files land in the wrong place."""
+    coro, ctx = deploy_one("assets/css/main file.css")
+    await coro
```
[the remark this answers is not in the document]
Shubhank-Jonnada: Captured the awaited result and asserted ACTION so the statement is no longer a bare await, which also means a failed deploy now reports itself instead of surfacing as a confusing assertion on the request list - https://github.com/Autohive-AI/autohive-integrations/commit/cfe402cd79fa14f845b9b9b8ceb82a273f18374b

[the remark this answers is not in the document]
Shubhank-Jonnada: Fixed in cfe402c

TheRealAgentK (approved):
This is good to go now, @Shubhank-Jonnada @JonoHive

PR #444, opened 2026-08-19
feat(karakeep): add self-hosted bookmark and tag integration

lohitya:
## Summary
This PR adds a Karakeep integration  so Autohive can save bookmarks and manage tags on **Karakeep Cloud** or a **self-hosted** instance. Same API either way: user supplies an HTTPS instance origin plus an API key.

## Authentication
Type: Custom (instance URL + API key)

Fields:
- `base_url` — HTTPS origin only (e.g. `https://cloud.karakeep.app` or `https://karakeep.example.com`). Trailing slash and `/api/v1` are not included. Non-origin values (paths, query strings, fragments, userinfo) are rejected. `http://` is rejected so the key is never sent in cleartext. Self-hosted Karakeep must sit behind TLS (reverse proxy or tunnel).
- `api_key` — created in the Karakeep app under **Settings → API Keys**. Sent as `Authorization: Bearer`. Grant bookmark and tag read/write scopes; admin scopes are not required.

Both fields are required in `auth.fields` so the SDK rejects incomplete auth before action execution.

## Actions

| Action | Behaviour |
|--------|-----------|
| `create_bookmark` | POST a bookmark of type `link` (default), `text`, or `asset`. Optional title, note, summary, archived, favourited. For **link** bookmarks, HTTP 201 = new; HTTP 200 = that URL already existed (`already_existed=true`, same id). Dedup does not apply to text/asset. Asset bookmarks need an `asset_id` from a prior Karakeep `POST /assets` — this integration does not upload files. |
| `attach_tags` | POST tags by **name** and/or existing **id**. Missing names are created. Optional `attached_by` (`ai` / `human`); omitted unless set (Karakeep defaults to `human`). |
| `search_bookmarks` | GET search (`q`), paginated. Optional `search_mode` (`fts` / `semantic` / `hybrid`), `sort_order`, `include_content`. |
| `get_bookmark` | GET one bookmark by id. Optional `include_content`. |
| `create_tag` | POST a tag by name. |
| `list_tags` | GET tags, optional `name_contains`, `sort`, `attached_by`, paginated (max 1000). |
| `list_bookmarks` | GET bookmarks, optional archived/favourited filters, `sort_order`, `include_content`, paginated (max 100). |
| `get_tag_bookmarks` | GET bookmarks for a tag id, optional `sort_order` / `include_content`, paginated. |

Boolean query params are sent as `"true"` / `"false"` strings. Path ids are URL-encoded.

## Validation
- **Unit (CI):** `karakeep/tests/test_karakeep_unit.py` — **58 passed**. Mocked: HTTPS-only and origin-only `base_url`; missing auth → `VALIDATION_ERROR`; create link/text/asset + type-specific required fields; link dedup 200 vs 201; attach by name/id/`attached_by`; search/list query params and limit caps; path-id encoding; fetch errors do not leak the API key.
- **Live (not CI):** `karakeep/tests/test_karakeep_integration.py`. Read-only vs destructive (`pytest.mark.destructive`). Needs `KARAKEEP_BASE_URL` (https only) + `KARAKEEP_API_KEY` (documented in root `.env.example`). Destructive lifecycle creates tag + bookmark, verifies attach/get/dedup, then deletes them in `finally`.
- **Tooling:** `validate_integration.py`, `check_config_sync.py`, `check_fetch_pattern.py`, `check_imports.py`, `check_readme.py`, `check_version_bump.py` (new integration `1.0.0`), ruff check/format, bandit.

## Running tests

```bash
# Unit
uv run --with-requirements requirements-test.txt --with-requirements karakeep/requirements.txt \
  pytest karakeep/tests/test_karakeep_unit.py

# Live, read-only
uv run --with-requirements requirements-test.txt --with-requirements karakeep/requirements.txt \
  pytest karakeep/tests/test_karakeep_integration.py -m "integration and not destructive"

# Live including destructive (creates then deletes a real tag + bookmark)
uv run --with-requirements requirements-test.txt --with-requirements karakeep/requirements.txt \
  pytest karakeep/tests/test_karakeep_integration.py -m "integration and destructive"

### CODE REVIEW

File: karakeep/tests/test_karakeep_integration.py [hunk 1]
```diff
@@ -0,0 +1,161 @@
...
+
+        by_tag = _assert_ok(
+            await karakeep.execute_action(
+                "get_tag_bookmarks",
+                {"tag_id": tag["tag_id"], "limit": 10},
+                live_context,
+            )
+        )
+        ids = [b.get("id") for b in by_tag.get("bookmarks") or []]
+        assert bookmark_id in ids
```
TheRealAgentK: Could this lifecycle clean up the bookmark and tag in a `finally` block? It currently leaves permanent test data on every run. Using only the newly created tag in `attach_tags` would also avoid creating the additional `source:autohive-integration-test` tag that cannot be safely deleted if it already existed.

karakeep/config.json
```diff
@@ -0,0 +1,424 @@
...
+                    "label": "Karakeep URL",
+                    "help_text": "Karakeep instance URL (https:// required). For Karakeep Cloud, use https://cloud.karakeep.app. For self-hosted, use your HTTPS URL such as https://karakeep.example.com. Karakeep does not terminate TLS itself — front your self-hosted instance with a reverse proxy or tunnel (Cloudflare Tunnel, Tailscale Funnel, Caddy, nginx, etc.). http:// is not accepted. Do not add a trailing slash or /api/v1."
+                },
+                "api_key": {
+                    "type": "string",
+                    "format": "password",
+                    "label": "API Key",
+                    "help_text": "API key from your Karakeep instance (Settings \u2192 API Keys). When creating the key, grant scopes for bookmarks and tags (read + write). Do not grant admin scopes unless needed."
+                }
+            }
```
TheRealAgentK: Please add `"required": ["base_url", "api_key"]` to `auth.fields`. Both credentials are mandatory in every handler, so the SDK should reject an incomplete auth object before action execution.

karakeep/karakeep.py
```diff
@@ -0,0 +1,376 @@
...
+        raise ValueError("api_key is required in auth")
+    return api_key
+
+
+def get_base_url(context: ExecutionContext) -> str:
+    base_url = str(_credentials(context).get("base_url") or "").strip().rstrip("/")
+    if not base_url:
+        raise ValueError("base_url is required in auth")
+    parsed = urlparse(base_url)
+    if parsed.scheme != "https" or not parsed.netloc:
```
TheRealAgentK, line 38: Since this field is documented as an HTTPS origin, consider rejecting non-root paths, query strings, fragments, and userinfo here. Otherwise a value such as `https://host/api/v1` passes validation and produces a duplicated `/api/v1/api/v1/...` path.

karakeep/README.md
```diff
@@ -0,0 +1,276 @@
...
+Plain `http://` URLs are refused — the API key is sent on every request.
+
+```bash
+# Read-only integration tests (default — safe to run against any instance)
+uv run --with-requirements requirements-test.txt --with-requirements karakeep/requirements.txt \
+  pytest karakeep/tests/test_karakeep_integration.py -m "integration and not destructive"
+
+# Include destructive tests (creates and modifies real data)
+uv run --with-requirements requirements-test.txt --with-requirements karakeep/requirements.txt \
+  pytest karakeep/tests/test_karakeep_integration.py -m integration
```
TheRealAgentK: Could this use `-m "integration and destructive"`? That keeps the opt-in command focused on mutating tests instead of rerunning the read-only suite too.

TheRealAgentK (changes requested):

karakeep/config.json
TheRealAgentK: Just tagging in @NinosMan - we did fix the issue with required auth fields in SDK 2.0.1, correct?

lohitya:
> Could you share the pytest output from running the full lifecycle E2E test against your actual Karakeep instance? Just the command output showing the live test passes is enough.

image

karakeep/tests/test_karakeep_integration.py
lohitya: Makes sense. I skipped cleanup at first so I could see the tag/bookmark in the UI while testing locally, but I agree runs shouldn’t leave permanent data.

karakeep/config.json
TheRealAgentK: Just chatted with @NinosMan and we're both very confident that this should work fine with 2.0.1

TheRealAgentK:
@lohitya Kai opened a small follow-up PR against your branch to harden the Karakeep lifecycle test cleanup: https://github.com/lohitya/autohive-integrations/pull/1

It uses UUID-based resource names, only deletes bookmarks confirmed as newly created, and surfaces cleanup failures while still attempting both deletions. Local code checks and all 58 unit tests pass.

TheRealAgentK (approved):
LGTM. Thx @lohitya for acting on the feedback!

